PR #2, opened 2017-04-21
make working directories for example configurations.

lcy-seso:
initialize working directories for examples.


reyoung (approved):
LGTM, Excellent job.

PR #17, opened 2017-05-02
add rnn train

chengxiaohua1105:
In this example, I explain how to train a language model by RNN.

### DIFF
--- a/language_model/rnn_train.py
+++ b/language_model/rnn_train.py
@@ -1,0 +1,104 @@
+import sys
+import sqlite3
+import paddle.v2 as paddle
+import ptb
+import gzip
+
+
+def rnn_net(dict_dim, is_generating=False):
+    # dimension of word vector
+    word_vector_dim = 1024
+    #dimension of hidden unit in GRU Encoder network
+    encoder_size = 1024
+
+    # Encoder
+    src_word_id = paddle.layer.data(
+        name='source_word',
+        type=paddle.data_type.integer_value_sequence(dict_dim))
+    target_word_id = paddle.layer.data(
+        name='target_word',
+        type=paddle.data_type.integer_value_sequence(dict_dim))
+
+    src_embedding = paddle.layer.embedding(
+        input=src_word_id,
+        size=word_vector_dim,
+        param_attr=paddle.attr.ParamAttr(name='_source_word_embedding'))
+
+    src_forward = paddle.networks.simple_gru(
+        input=src_embedding, size=encoder_size)
+    src_backward = paddle.networks.simple_gru(
+        input=src_embedding, size=encoder_size, reverse=True)
+    encoded_vector = paddle.layer.concat(input=[src_forward, src_backward])
+    output = paddle.layer.fc(
+        input=encoded_vector,
+        size=dict_dim,
+        act=paddle.activation.Softmax(),
+        param_attr=paddle.attr.Param(initial_std=0.01))
+    # calc cost
+    cost = paddle.layer.classification_cost(input=output, label=target_word_id)
+
+    return cost
+
+
+def main():
+    paddle.init(use_gpu=False, trainer_count=8)
+    is_generating = False
+
+    # source and target dict dim.
+    word_dict = ptb.build_dict()
+    dict_size = len(word_dict)
+    # train the network
+    if not is_generating:
+        cost = rnn_net(dict_size)
+        parameters = paddle.parameters.create(cost)
+
+        # define optimize method and trainer
+        optimizer = paddle.optimizer.Adam(
+            learning_rate=5e-3,
+            regularization=paddle.optimizer.L2Regularization(rate=8e-4))
+
+        trainer = paddle.trainer.SGD(
+            cost=cost, parameters=parameters, update_equation=optimizer)
+
+        # define data reader
+        ptb_reader = paddle.batch(
+            paddle.reader.shuffle(ptb.seq_train(word_dict, 100), buf_size=8192),
+            batch_size=64)
+
+        # define event_handler callback
+        def event_handler(event):
+            # save model
+            if isinstance(event, paddle.event.EndPass):
+                # need make directory models in current dir before running
+                model_name = './models/model_pass_%05d.tar.gz' % event.pass_id
+                print("Save model into %s ..." % model_name)
+                with gzip.open(model_name, 'w') as f:
+                    parameters.to_tar(f)
+
+            # output training info
+            if isinstance(event, paddle.event.EndIteration):
+                if event.batch_id % 10 == 0:
+                    result = trainer.test(
+                        paddle.batch(
+                            paddle.reader.shuffle(
+                                ptb.seq_test(word_dict, 100), buf_size=8192),
+                            batch_size=64))
+                    print "\nPass %d, Batch %d, Cost %f, %s, Test Cost %f, %s"\
+                        % (
+                            event.pass_id, event.batch_id, event.cost,
+                            event.metrics, result.cost, result.metrics)
+                else:
+                    sys.stdout.write('.')
+                    sys.stdout.flush()
+
+        # start to train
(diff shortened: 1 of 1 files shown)


language_model/rnn_train.py
```diff
@@ -0,0 +1,104 @@
...
+        param_attr=paddle.attr.Param(initial_std=0.01))
+    # calc cost
+    cost = paddle.layer.classification_cost(input=output, label=target_word_id)
+
+    return cost
+
+
+def main():
+    paddle.init(use_gpu=False, trainer_count=8)
+    is_generating = False
```
wangkuiyi: I'd thought that we don't need `is_generating` in programs written using v2 API.

language_model/rnn_train.py
```diff
@@ -0,0 +1,104 @@
+import sys
+import sqlite3
+import paddle.v2 as paddle
+import ptb
+import gzip
+
+
+def rnn_net(dict_dim, is_generating=False):
```
wangkuiyi, line 8: `is_generating` is not used in this function.

language_model/rnn_train.py
```diff
@@ -0,0 +1,104 @@
...
+        # define optimize method and trainer
+        optimizer = paddle.optimizer.Adam(
+            learning_rate=5e-3,
+            regularization=paddle.optimizer.L2Regularization(rate=8e-4))
+
+        trainer = paddle.trainer.SGD(
+            cost=cost, parameters=parameters, update_equation=optimizer)
+
+        # define data reader
+        ptb_reader = paddle.batch(
```
wangkuiyi, line 64: @helinwang I'd thought that `paddle.batch` is supposed to be called by `Trainer.train` and is hidden from users. But I just checked that `paddle.batch` is implemented in `minibatch.py` and exported to users via `__init__.py`.  Is this inconsistent with our design?

wangkuiyi (commented):
Awesomely rapid PR!

Is this program runnable?

language_model/rnn_train.py
helinwang: @wangkuiyi Yes it's consistent (see [here](https://github.com/PaddlePaddle/Paddle/blob/develop/doc/design/reader/README.md#usage)). We agreed on the design because the user may want to implement the batch reader directly (e.g., some nlp task require data in a mini-batch be sorted according to the length of each data instance), so we allow the user to pass in the batch reader.
However, the more I think about it, the more I wish to let `train` only accept reader (currently train only accept `batch reader`, not `reader`). We can specify the batch size as an argument, so the trainer will use `paddle.batch` internally. Or if user wants to do some customization (such as sorting), they can pass in a batcher to do it.

language_model/rnn_train.py
```diff
@@ -0,0 +1,104 @@
+import sys
+import sqlite3
```
lcy-seso, line 2: The package sqlite3 is not explicitly used in this script, so it should not be imported.

language_model/rnn_train.py
```diff
@@ -0,0 +1,104 @@
...
+import paddle.v2 as paddle
+import ptb
```
lcy-seso, line 4: ptb is not used in this script, so it should not be imported, or users who do not install this package cannot run this script.

language_model/rnn_train.py
lcy-seso: `is_generating` is not used and it will not be a parameter for this function. I think it should be removed.

In the `language model` example, a generating is required and is expected to show how to generate a sequence by using a trained language model, which is quite a different from training, but generating is not implemented yet in this pr.

1. the encoding RNN: load the trained model to encode the given input sequence and output the sentence encoding (a dense vector) of the input sentence.
2. the generate RNN: initialized by the sentence encoding calculated by the first part RNN, every time, it takes the most probable words in last time step as inputs.

Using a trained RNN language model to generate a sequence is not straightforward. It begins by given an input sentence, and then generate the following sentence. You should split the original trained RNN model into two parts, which have the same architecture and share the trained parameter, but have different effects: the first part is to encode the input sequence, while the second part is to generate.

To generate a sequence, you have to implement the OneWay search or BeamSearch (optional) algorithm. I recommend using `recurrent_layer_group` to implement the RNN language model to avoid implementing OneWay search or BeamSearch yourself. Currently, in PaddlePaddle, generating algorithms like OneWay search or BeamSearch is only implemented in `recurrent_layer_group`.

language_model/rnn_train.py
```diff
@@ -0,0 +1,104 @@
...
+        optimizer = paddle.optimizer.Adam(
+            learning_rate=5e-3,
+            regularization=paddle.optimizer.L2Regularization(rate=8e-4))
+
+        trainer = paddle.trainer.SGD(
+            cost=cost, parameters=parameters, update_equation=optimizer)
+
+        # define data reader
+        ptb_reader = paddle.batch(
+            paddle.reader.shuffle(ptb.seq_train(word_dict, 100), buf_size=8192),
```
lcy-seso, line 65: I wonder whether this script can run or not. I am confusing with the input data format and how to provide data to PaddlePaddle models from the raw input data. 

We cannot provide an example that cannot run or an example that we cannot explain how to run to the users. 

Is the `ptb`  module a dataset you want to add into paddle.dataset for `the Penn Treebank` dataset?  If so, it should be put into paddle.dataset. The name `ptb` is very confusing with the python  traceback module.

language_model/rnn_train.py
```diff
@@ -0,0 +1,104 @@
...
+                with gzip.open(model_name, 'w') as f:
+                    parameters.to_tar(f)
+
+            # output training info
+            if isinstance(event, paddle.event.EndIteration):
+                if event.batch_id % 10 == 0:
+                    result = trainer.test(
+                        paddle.batch(
+                            paddle.reader.shuffle(
+                                ptb.seq_test(word_dict, 100), buf_size=8192),
```
lcy-seso, line 84: More details about training and testing dataset should be added.

The dataset `ptb` is not implemented now.

lcy-seso (commented):
I think this example should be rewritten by using recurrent_layer_group, otherwise, it cannot be used in sequence generation.

language_model/rnn_train.py
pkuyym: @lcy-seso I implement **ptb** based on **imikolov**，will push in next pull request.

language_model/rnn_train.py
chengxiaohua1105: yes it can run when import ptb which is implemented by @pkuyym . It's just training for language model, and  I will implement how to generate later.

chengxiaohua1105: is_generating is a variable used for generating

lcy-seso:
I think there are two main problems in this first PR:
+ RNN language model should be rewritten by `recurrent_layer_group`, or it cannot be used to generate a sequence.
+ I do not think this PR can run. More details about the raw data and how to prepare data are needed.

language_model/rnn_train.py
```diff
@@ -0,0 +1,104 @@
...
+
+    src_embedding = paddle.layer.embedding(
+        input=src_word_id,
+        size=word_vector_dim,
+        param_attr=paddle.attr.ParamAttr(name='_source_word_embedding'))
+
+    src_forward = paddle.networks.simple_gru(
+        input=src_embedding, size=encoder_size)
+    src_backward = paddle.networks.simple_gru(
+        input=src_embedding, size=encoder_size, reverse=True)
```
lcy-seso, line 30: The bidirectional RNN is not necessary and it should be removed. The most serious problem is that the bidirectional RNN cannot be used in sequence generation.

lcy-seso (commented):
The bidirectional RNN should not be used because it cannot be used in sequence generation.

lcy-seso (changes requested):
Request changes.

lcy-seso:
Please merge the develop branch to enable travis-CI.

PR #72, opened 2017-06-05
Add loading model function for train.py.

### LINKED ISSUE #79: Add loading model function for train.py

pkuyym:
fix #79

### DIFF
--- a/deep_speech_2/train.py
+++ b/deep_speech_2/train.py
@@ -11,6 +11,7 @@ import sys
 from model import deep_speech2
 from audio_data_utils import DataGenerator
 import numpy as np
+import os
 
 #TODO: add WER metric
 
@@ -78,6 +79,11 @@ parser.add_argument(
     default='data/eng_vocab.txt',
     type=str,
     help="Vocabulary filepath. (default: %(default)s)")
+parser.add_argument(
+    "--init_model_path",
+    default='models/params.tar.gz',
+    type=str,
+    help="Model path for initialization. (default: %(default)s)")
 args = parser.parse_args()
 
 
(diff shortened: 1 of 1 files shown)


deep_speech_2/train.py
```diff
@@ -78,6 +79,11 @@
     default='data/eng_vocab.txt',
     type=str,
     help="Vocabulary filepath. (default: %(default)s)")
+parser.add_argument(
+    "--init_model_path",
+    default='models/params.tar.gz',
```
xinghai-sun: Set default  to None (training from scratch).

deep_speech_2/train.py
```diff
@@ -78,6 +79,11 @@
...
+    type=str,
+    help="Model path for initialization. (default: %(default)s)")
```
xinghai-sun: Add e.g. "If set None, the training will start from scratch. Otherwise, the training will resume from the existing model of this path".

deep_speech_2/train.py
```diff
@@ -114,8 +120,13 @@ def train():
         rnn_size=args.rnn_layer_size,
         is_inference=False)
 
-    # create parameters and optimizer
-    parameters = paddle.parameters.create(cost)
+    # create/load parameters and optimizer
+    if args.init_model_path is None:
+        parameters = paddle.parameters.create(cost)
+    else:
+        assert os.path.isfile(args.init_model_path), "Invalid model."
```
xinghai-sun: It would be better to use "raise IOError" ?

xinghai-sun (changes requested):
Good. Thanks.

deep_speech_2/train.py
pkuyym: Done.

deep_speech_2/train.py
pkuyym: Done.

deep_speech_2/train.py
pkuyym: Agree and done.

xinghai-sun (approved):
LGTM

PR #88, opened 2017-06-13
Expand .travis.yml to support docker.

### LINKED ISSUE #87: Add unittest for ci
There are several new python modules added in deep speech project. These modules should be tested before merging. Since some scripts need PaddlePaddle running environment, so I expand the ```.travis.yml``` to support docker. There are several rules when writing unit test scripts.
1. all unit test scripts should be placed into **tests** directory.
2. all unit test scripts should be named like ```test*.py```.

pkuyym:
resolves #87

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: cpp
 cache: ccache
 sudo: required
 dist: trusty
+services:
+  - docker
 os:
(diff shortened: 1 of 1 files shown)


gangliao (approved):
LGTM

Issue #100, opened 2017-06-15
Abnormal learning curve bumping at early batches of each epoch during DS2 training.


xinghai-sun:
After merging PR #74, we have seen such abnormal learning curve:

<img width="500" alt="7eb432e75966334978e6adb2b" src="https://user-images.githubusercontent.com/7038341/27165435-7dcfe1a2-51c6-11e7-8b18-5660da44183a.png">

The figure plots the training cost. Notice that in the tails of the curve, there are many spikes, exactly locating at the first batch of each epoch. 

Besides, it is not easy to reproduce the phenomenon in a small dataset.

xinghai-sun:
One weird thing: when I resume the training from a saved model from the above figure, the phenomena did not appear again.

qingqing01:
抱歉，我用中文吧 :(

基于`libri.train-clean-100`，14个pass开始收敛比较奇怪，突然上涨，然后下降：

```
Pass: 13, Batch: 800, TrainCost: 33.875851
.................................................
Pass: 13, Batch: 850, TrainCost: 32.388756
.........................................
------- Time: 2996 sec,  Pass: 13, ValidationCost: 270.575763434

Pass: 14, Batch: 0, TrainCost: 45.968803
.................................................
Pass: 14, Batch: 50, TrainCost: 492.662450

```

去掉batch-shuffle中的下面几行，即扔掉开头一些短样本，和不够组batch的长样本，

```
res_len = len(manifest) - shift_len - len(batch_manifest)
batch_manifest.extend(manifest[-res_len:])
batch_manifest.extend(manifest[0:shift_len])
```

收敛情况没有出现突然上升，看着都比较正常：

```
.........

Pass: 19, Batch: 890, TrainCost: 25.492654 CurCost: 14.879614

------- Time: 2977 sec,  Pass: 19, ValidationCost: 61.139245818

Pass: 20, Batch: 0, TrainCost: 27.037848 CurCost: 27.037848


```

xinghai-sun:
I've given up the attempt to reproduce the phenomenon from a pre-trained model. 

Now I've started three from-scratch jobs with three different shuffle methods, i.e. 
- **instance shuffle**
- **batch shuffle**
- **batch shuffle with clipping**

(For more details, please refer [here](https://github.com/PaddlePaddle/models/pull/101))

with full LibriSpeech data, in order to reproduce what @qingqing01  has observed in a small dataset.

xinghai-sun:
[...]

PR #101, opened 2017-06-15
Add shuffle type of instance_shuffle and batch_shuffle_clipped.

xinghai-sun:
1. Add shuffle type of `instance_shuffle` and `batch_shuffle_clipped` to help investigate problems like #100.
2. Enable `min_batch_size` to temporarily avoid "batch_size < trainer_count" error (A permanent fix should be done on Paddle).
3. Add argument print.


deep_speech_2/train.py
```diff
@@ -52,6 +53,12 @@
     type=distutils.util.strtobool,
     help="Use sortagrad or not. (default: %(default)s)")
 parser.add_argument(
+    "--shuffle_method",
+    default='instance_shuffle',
```
qingqing01, line 57: We should choose the best default mode after experiments.

qingqing01 (approved):

deep_speech_2/train.py
xinghai-sun: Yes. Will do that after we get a conclusion.

PR #111, opened 2017-06-20
Add multi-threading support for DS2 data generator.

### LINKED ISSUE #110: Add multi-threading support for DS2 data generator.
Add multi-threading support for DS2 data generator, to accelerate the training.

xinghai-sun:
resolve #110 

No difference in the generated data, but with 6x speedup (reduce ~1.5 hours per pass).

### DIFF
--- a/deep_speech_2/data_utils/data.py
+++ b/deep_speech_2/data_utils/data.py
@@ -44,6 +44,8 @@ class DataGenerator(object):
     :types max_freq: None|float
     :param specgram_type: Specgram feature type. Options: 'linear'.
     :type specgram_type: str
+    :param num_threads: Number of CPU threads for processing data.
+    :type num_threads: int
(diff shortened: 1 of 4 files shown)


deep_speech_2/data_utils/data.py
```diff
@@ -58,6 +60,7 @@ def __init__(self,
                  window_ms=20.0,
                  max_freq=None,
                  specgram_type='linear',
+                 num_threads=12,
```
wanghaoshuang: The default value of num_threads  can be setted to [multiprocessing.cpu_count()](https://docs.python.org/2/library/multiprocessing.html#multiprocessing.cpu_count)?

xinghai-sun: Done. Thanks.

wanghaoshuang (approved):

PR #173, opened 2017-07-19
"add start script"

dzhwinter:


ltr/lambda_rank.py
```diff
@@ -117,6 +118,15 @@ def lambda_rank_infer(pass_id):
...
+        type=int,
+        help="num of passes in train| infer pass number of model")
+    args = parser.parse_args()
     paddle.init(use_gpu=False, trainer_count=1)
-    train_lambda_rank(2)
-    lambda_rank_infer(pass_id=1)
+    if args.run_type == "train":
+        train_lambda_rank(args.num_passes)
+    elif args.run_type == "infer":
+        lambda_rank_infer(pass_id=args.pass_num - 1)
```
typhoonzero, line 132: Should select best pass? The last pass may not be best?

ltr/run_lambdarank.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+python lambda_rank.py \
```
typhoonzero, line 3: Can add `stdbuf -oL xxxx` to line-buffer the pipe.

ltr/lambda_rank.py
dzhwinter: demo case do not ensure the best model is selected. Just a demo show. : )

ltr/run_lambdarank.sh
dzhwinter: 目测不需要缓冲，只是demo show，展示功能正常。这个start script也是为了方便小白用户使用的

typhoonzero (approved):
LGTM!

livc:
这个pr对应model的文档也要更新吧？

PR #183, opened 2017-08-01
Simplify train.py, evaluate.py, infer.py and tune.py by adding DeepSpeech2Model class for DS2.

xinghai-sun:
1. Move functions in model.py into layer.py
2. Add a DeepSpeech2Model class in model.py, which has train() and infer_batch() methods.
3. Simplify train.py, evaluate.py, infer.py and tune.py。

### DIFF
--- a/deep_speech_2/evaluate.py
+++ b/deep_speech_2/evaluate.py
@@ -4,6 +4,7 @@ from __future__ import division
 from __future__ import print_function
 
 import distutils.util
+import sys
 import argparse
 import gzip
 import paddle.v2 as paddle
@@ -12,13 +13,19 @@ from model import deep_speech2
 from decoder import *
 from lm.lm_scorer import LmScorer
 from error_rate import wer
+import utils
 
 parser = argparse.ArgumentParser(description=__doc__)
 parser.add_argument(
     "--batch_size",
-    default=100,
+    default=128,
     type=int,
     help="Minibatch size for evaluation. (default: %(default)s)")
+parser.add_argument(
+    "--trainer_count",
+    default=8,
+    type=int,
+    help="Trainer number. (default: %(default)s)")
 parser.add_argument(
     "--num_conv_layers",
     default=2,
@@ -58,8 +65,8 @@ parser.add_argument(
     "--decode_method",
     default='beam_search',
     type=str,
-    help="Method for ctc decoding, best_path or beam_search. (default: %(default)s)"
-)
+    help="Method for ctc decoding, best_path or beam_search. "
+    "(default: %(default)s)")
 parser.add_argument(
     "--language_model_path",
     default="lm/data/common_crawl_00.prune01111.trie.klm",
@@ -67,12 +74,12 @@ parser.add_argument(
     help="Path for language model. (default: %(default)s)")
 parser.add_argument(
     "--alpha",
-    default=0.26,
+    default=0.36,
     type=float,
     help="Parameter associated with language model. (default: %(default)f)")
 parser.add_argument(
     "--beta",
-    default=0.1,
+    default=0.25,
     type=float,
     help="Parameter associated with word count. (default: %(default)f)")
 parser.add_argument(
@@ -191,7 +198,7 @@ def evaluate():
                 blank_id=len(data_generator.vocab_list),
                 num_processes=args.num_processes_beam_search,
                 ext_scoring_func=ext_scorer,
-                cutoff_prob=args.cutoff_prob, )
+                cutoff_prob=args.cutoff_prob)
             for i, beam_search_result in enumerate(beam_search_results):
                 wer_sum += wer(target_transcription[i],
                                beam_search_result[0][1])
@@ -199,12 +206,15 @@ def evaluate():
         else:
             raise ValueError("Decoding method [%s] is not supported." %
                              decode_method)
+        print("WER (%d/?) = %f" % (wer_counter, wer_sum / wer_counter))
 
-    print("Final WER = %f" % (wer_sum / wer_counter))
+    print("Final WER (%d/%d) = %f" % (wer_counter, wer_counter,
+                                      wer_sum / wer_counter))
 
 
 def main():
-    paddle.init(use_gpu=args.use_gpu, trainer_count=1)
+    utils.print_arguments(args)
+    paddle.init(use_gpu=args.use_gpu, trainer_count=args.trainer_count)
     evaluate()
 
 
--- a/deep_speech_2/evaluate.py
+++ b/deep_speech_2/evaluate.py
@@ -4,14 +4,11 @@ from __future__ import division
 from __future__ import print_function
 
 import distutils.util
-import sys
 import argparse
-import gzip
+import multiprocessing
 import paddle.v2 as paddle
 from data_utils.data import DataGenerator
-from model import deep_speech2
-from decoder import *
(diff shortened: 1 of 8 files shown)


deep_speech_2/tune.py
```diff
@@ -62,10 +66,10 @@
     type=str,
     help="Manifest path for normalizer. (default: %(default)s)")
 parser.add_argument(
-    "--decode_manifest_path",
+    "--tune_manifest_path",
     default='datasets/manifest.test',
```
kuke: ```manifest.dev``` would be better.

deep_speech_2/layer.py
```diff
@@ -0,0 +1,155 @@
...
+        conv = conv_bn_layer(
+            input=conv,
+            filter_size=(11, 21),
+            num_channels_in=32,
+            num_channels_out=32,
+            stride=(1, 2),
+            padding=(5, 10),
+            act=paddle.activation.BRelu())
+    output_num_channels = 32
+    output_height = 160 // pow(2, num_stacks) + 1
```
kuke, line 99: Can we figure out a way to avoid hardcode here?

deep_speech_2/train.py
```diff
@@ -127,100 +129,47 @@
```
kuke, line 134: Is it necessary to add ```output_model_dir``` into ```parser``` as an argument?

deep_speech_2/evaluate.py
```diff
@@ -112,99 +116,52 @@
...
-                wer_counter += 1
-        else:
-            raise ValueError("Decoding method [%s] is not supported." %
-                             decode_method)
-
-    print("Final WER = %f" % (wer_sum / wer_counter))
+        for target, result in zip(target_transcripts, result_transcripts):
+            wer_sum += wer(target, result)
+            num_ins += 1
+        print("WER (%d/?) = %f" % (num_ins, wer_sum / num_ins))
```
pkuyym, line 158: what does ```?``` mean?

deep_speech_2/layer.py
```diff
@@ -0,0 +1,155 @@
...
+    forward_simple_rnn = paddle.layer.recurrent(
+        input=input_proj_bn, act=act, reverse=False)
+    backward_simple_rnn = paddle.layer.recurrent(
+        input=input_proj_bn, act=act, reverse=True)
+    return paddle.layer.concat(input=[forward_simple_rnn, backward_simple_rnn])
+
+
+def conv_group(input, num_stacks):
+    """
+    Convolution group with several stacking convolution layers.
```
kuke: Seems that ```stacked``` is often used instead of ```stacking```. The same below.

deep_speech_2/layer.py
pkuyym, line 99: Since the model is refactored, please make ```160``` exposed instead of hard-coding.

kuke (commented):
Great! Almost LGTM.

deep_speech_2/layer.py
```diff
@@ -0,0 +1,155 @@
+"""Contains DeepSpeech2 layers."""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import paddle.v2 as paddle
+
+DISABLE_CUDNN_BATCH_NORM = True
```
pkuyym: Why not make ```DISABLE_CUDNN_BATCH_NORM``` a parameter of ```conv_bn_layer```? If hard coding here, user has to modify this file when training on ```cpu``` mode.

deep_speech_2/model.py
```diff
@@ -3,141 +3,150 @@
...
+                    sys.stdout.flush()
+            if isinstance(event, paddle.event.BeginPass):
+                start_time = time.time()
+                cost_sum, cost_counter = 0.0, 0
+            if isinstance(event, paddle.event.EndPass):
+                result = trainer.test(
+                    reader=dev_batch_reader, feeding=feeding_dict)
+                output_model_path = os.path.join(
+                    output_model_dir, "params.pass-%d.tar.gz" % event.pass_id)
+                with gzip.open(output_model_path, 'w') as f:
```
pkuyym, line 109: Maybe we should make sure ```output_model_dir``` exist, the saving operation may fail if the directory is not exist.

deep_speech_2/model.py
```diff
@@ -3,141 +3,150 @@
...
-            blank=dict_size,
-            norm_by_times=True)
+    def _create_network(self, vocab_size, num_conv_layers, num_rnn_layers,
+                        rnn_layer_size):
+        # paddle.data_type.dense_array is used for variable batch input.
+        # The size 161 * 161 is only an placeholder value and the real shape
+        # of input batch data will be induced during training.
+        audio_data = paddle.layer.data(
+            name="audio_spectrogram",
+            type=paddle.data_type.dense_array(161 * 161))
```
pkuyym, line 212: Just curious about ```161 * 161```, is this setting proper for other type of feature like ```mfcc```?

pkuyym (changes requested):

deep_speech_2/model.py
```diff
@@ -3,141 +3,150 @@
...
-        return paddle.layer.warp_ctc(
-            input=fc,
-            label=text_data,
-            size=dict_size + 1,
-            blank=dict_size,
-            norm_by_times=True)
+    def _create_network(self, vocab_size, num_conv_layers, num_rnn_layers,
+                        rnn_layer_size):
+        # paddle.data_type.dense_array is used for variable batch input.
+        # The size 161 * 161 is only an placeholder value and the real shape
```
kuke: The log information about output dimensions of conv layers is induced  from here, which is misleading if this placeholder is not consistent with the real shape. Maybe a better choice is to use the real dimension of feature vector.

deep_speech_2/model.py
xinghai-sun: Done.

deep_speech_2/model.py
xinghai-sun: Will fix this in another PR.

deep_speech_2/model.py
xinghai-sun: Will fix this in another PR. Lets' just do not bring in any difference to this PR.

deep_speech_2/train.py
xinghai-sun: Done.

deep_speech_2/tune.py
xinghai-sun: Done.

deep_speech_2/evaluate.py
xinghai-sun: It refers to an unknown size of the validation set. It will become known at the end of the evaluation.

deep_speech_2/layer.py
xinghai-sun: Removed since the cudnn problem has been fixed.

deep_speech_2/layer.py
xinghai-sun: Done.

deep_speech_2/layer.py
xinghai-sun: Will be fixed in a later PR (this problem is not coming from current PR).

xinghai-sun: Will be fixed in a later PR (this problem is not coming from current PR).

pkuyym (approved):
LGTM

PR #186, opened 2017-08-03
Add a realtime ASR demo (both server and client) for DS2 users to try with own voice.

xinghai-sun:


deep_speech_2/README.md
```diff
@@ -143,3 +143,19 @@ python tune.py --help
 ```
 
 Then reset parameters with the tuning result before inference or evaluating.
+
+### Playing with the ASR Demo
+
+A real-time ASR demo (`demo_server.py` and `demo_client.py`) are prepared for users to try out the ASR model with their own voice. After a model and language model is prepared, we can first start the demo server:
```
luotao1: is prepared -> are prepared

deep_speech_2/data_utils/data.py
```diff
@@ -83,6 +83,23 @@ def __init__(self,
         self._rng = random.Random(random_seed)
         self._epoch = 0
 
+    def process_utterance(self, filename, transcript):
+        """Load, augment, featurize and normalize for speech data.
```
luotao1, line 87: """换一行再写注释，下同

deep_speech_2/demo_client.py
```diff
@@ -0,0 +1,94 @@
+"""Client-end for the ASR demo."""
```
luotao1, line 1: 开头这里用 # 来注释，可参考[layers.py](https://github.com/PaddlePaddle/Paddle/blob/develop/python/paddle/trainer_config_helpers/layers.py)的注释方式。下同。

deep_speech_2/README.md
```diff
@@ -16,6 +16,19 @@ export LD_LIBRARY_PATH=$PADDLE_INSTALL_DIR/Paddle/third_party/install/warpctc/li
 
 Please replace `$PADDLE_INSTALL_DIR` with your own paddle installation directory.
 
+### Setup for Demo
+
+Please do the following extra installation before run `demo_client.py` to try the realtime ASR demo. However there is no need to install them for the computer running the demo's server-end (`demo_server.py`). For details of running the ASR demo, please refer to the [section](#playing-with-the-asr-demo).
```
pkuyym: run --> running
I think it's better to let ASR demo be a single section including setup and running instructions.

deep_speech_2/model.py
```diff
@@ -35,6 +35,7 @@ def __init__(self, vocab_size, num_conv_layers, num_rnn_layers,
                              rnn_layer_size)
         self._create_parameters(pretrained_model_path)
         self._inferer = None
+        self._loss_inferer = None
```
pkuyym, line 38: ```self._cost_inferer``` be better?

deep_speech_2/model.py
```diff
@@ -118,6 +119,24 @@ def event_handler(event):
             num_passes=num_passes,
             feeding=feeding_dict)
 
+    def infer_loss_batch(self, infer_data):
```
pkuyym, line 122: ```infer_batch_cost``` be better?

deep_speech_2/model.py
```diff
@@ -118,6 +119,24 @@ def event_handler(event):
...
+        """Model inference. Infer the ctc loss for a batch of speech
+        utterances.
+
+        :param infer_data: List of utterances to infer, with each utterance a
```
pkuyym, line 126: with each utterance --> each utterance consists of

deep_speech_2/README.md
```diff
@@ -143,3 +156,23 @@ python tune.py --help
...
+And then in another console, start the client:
+
+```
+python demo_client.py
+```
+On the client console, press and hold "white-space" key and start talking, then release the "white-space" key when you finish your speech. The decoding results (infered transcription) will be displayed.
+
+If you would like to start the server and the client in two machines. Please use `--host_ip` and `--host_port` to indicate the actual IP address and port, for both `demo_server.py` and `demo_client.py`.
+
+Notice that `demo_client.py` should be started in your local computer with microphone hardware, while `demo_server.py` can be started in any remote server as well as the same local computer. IP address and port should be properly set for server-client communication.
```
pkuyym: Maybe we should point out that accessing to remote server through the network should be make sured.

deep_speech_2/data_utils/data.py
```diff
@@ -83,6 +83,23 @@ def __init__(self,
         self._rng = random.Random(random_seed)
         self._epoch = 0
 
+    def process_utterance(self, filename, transcript):
```
pkuyym, line 86: transcript -> transcription, they have different meanings.

deep_speech_2/demo_client.py
```diff
@@ -0,0 +1,94 @@
...
+def callback(in_data, frame_count, time_info, status):
+    """Audio recorder's stream callback function."""
+    global data_list, is_recording, enable_trigger_record
+    if is_recording:
+        data_list.append(in_data)
+        enable_trigger_record = False
+    elif len(data_list) > 0:
+        # Connect to server and send data
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        sock.connect((args.host_ip, args.host_port))
```
pkuyym, line 58: I think it's better to connect at first and reusing the connection rather than connecting every time when sending messages.

deep_speech_2/demo_server.py
```diff
@@ -0,0 +1,245 @@
...
+    default=0.25,
+    type=float,
+    help="Parameter associated with word count. (default: %(default)f)")
+parser.add_argument(
+    "--cutoff_prob",
+    default=0.99,
+    type=float,
+    help="The cutoff probability of pruning"
+    "in beam search. (default: %(default)f)")
+args = parser.parse_args()
```
pkuyym, line 111: Too many duplicated arguments with ```infer.py``` and ```evaluate.py```, maybe we can refactor this part later. Mark here.

deep_speech_2/demo_server.py
```diff
@@ -0,0 +1,245 @@
...
+        SocketServer.TCPServer.__init__(
+            self, server_address, RequestHandlerClass, bind_and_activate=True)
+
+
+class AsrRequestHandler(SocketServer.BaseRequestHandler):
+    """The ASR request handler."""
+
+    def handle(self):
+        # receive data through TCP socket
+        chunk = self.request.recv(1024)
```
pkuyym, line 134: I think it's better to make ```1024``` be an optional argument.

pkuyym (changes requested):
Great!

deep_speech_2/README.md
xinghai-sun: Move this to demo section.

deep_speech_2/README.md
xinghai-sun: Done.

deep_speech_2/README.md
xinghai-sun: Done.

deep_speech_2/data_utils/data.py
xinghai-sun: Both of transcription and transcript could refer to the text contents of speech. I use "transcription" for docs and "transcript" for code variables (due to the shorter length).

deep_speech_2/data_utils/data.py
xinghai-sun: The practice of 
```
def function():
"""This is function doc.
"""
```
 follows the [Google Coding Style for Python](http://zh-google-styleguide.readthedocs.io/en/latest/google-python-styleguide/python_style_rules/). And we keep this style consistent throughout the whole DS2 project.

deep_speech_2/demo_client.py
xinghai-sun: In layers.py, # is only used for copyrights declaration.
For file's doc, use `"""` instead.

deep_speech_2/demo_client.py
xinghai-sun: Opening a connection only cost several milliseconds. Besides, opening an independent connection for each utterance simplifies the codes (otherwise, in the server side, we have to handle multiple utterances with while-loop in a single handle() call).

deep_speech_2/demo_server.py
xinghai-sun: Yes. Let's discuss it later.

deep_speech_2/demo_server.py
xinghai-sun: No need for that. Chunk size does not matter.

deep_speech_2/model.py
xinghai-sun: No big difference between loss and cost. I prefer loss for it is more commonly used.

deep_speech_2/model.py
xinghai-sun: No big difference between loss and cost. I prefer loss for it is more commonly used.

deep_speech_2/model.py
xinghai-sun: --> with each utterance consisting of ....

pkuyym (approved):
LGTM

PR #188, opened 2017-08-08
Reset default multi-thread/process number to half of cpu count() for speedup.

xinghai-sun:
Change default thread/process number from `multiprocessing.cpu_count()` to `multiprocessing.cpu_count() // 2` to reduce thread/process competition.


xinghai-sun:
Larger than 20% speedup for some machines.

pkuyym (approved):
LGTM

PR #189, opened 2017-08-08
Accelerate mfcc computation for DS2.

xinghai-sun:
1. Put np.transpose on small tensor instead of larger tensor.
2. Use np.concatenate directly on ndarray instead of converting it to list and back to ndarray again.

Speedup by 2x for MFCC computation.


kuke (approved):
LGTM

PR #192, opened 2017-08-09
Update mfcc computation in DS2

kuke:
Fix a bug in PR#[189](https://github.com/PaddlePaddle/models/pull/189) by transposing  the three features after all their computation is completed.


xinghai-sun (approved):
LGTM

PR #195, opened 2017-08-10
Update readme in DS2

kuke:
Remove the limitation of cuDNN version due to [the bug fix](https://github.com/PaddlePaddle/Paddle/pull/3309)

### DIFF
--- a/deep_speech_2/README.md
+++ b/deep_speech_2/README.md
@@ -5,7 +5,7 @@
 ### Prerequisites
 
  - **Python = 2.7** only supported;
- - **cuDNN >= 6.0** is required to utilize NVIDIA GPU platform in the installation of PaddlePaddle, and the **CUDA toolkit** with proper version suitable for cuDNN. The cuDNN library below 6.0 is found to yield a fatal error in batch normalization when handling utterances with long duration in inference.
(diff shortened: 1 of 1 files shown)


deep_speech_2/README.md
```diff
@@ -5,7 +5,7 @@
 ### Prerequisites
 
  - **Python = 2.7** only supported;
- - **cuDNN >= 6.0** is required to utilize NVIDIA GPU platform in the installation of PaddlePaddle, and the **CUDA toolkit** with proper version suitable for cuDNN. The cuDNN library below 6.0 is found to yield a fatal error in batch normalization when handling utterances with long duration in inference.
+ - **cuDNN** library is required to utilize NVIDIA GPU platform sufficiently in the installation of PaddlePaddle, and the **CUDA toolkit** with proper version suitable for cuDNN.
```
xinghai-sun: If Python 2.7 and cuDNN is already required when installing PaddlePaddle, we don't have to claim them again.

xinghai-sun (approved):
I recommend to remove the whole `Prerequisites`.

deep_speech_2/README.md
pkuyym: maybe we can specify some version for ```proper version```

kuke: Agree with that

PR #202, opened 2017-08-14
Rename self.local_data to self._local_data in class DataGenerator.

### LINKED ISSUE #203: Rename self.local_data to self._local_data in class DataGenerator.

wanghaoshuang:
fix #203


lcy-seso (approved):
LGTM.

PR #208, opened 2017-08-16
deep speech2 can directly use warpctc instead by export LD_LIBRARY_PATH

luotao1:
Inspired by https://github.com/PaddlePaddle/Paddle/pull/3503, we can directly use warpctc instead by export LD_LIBRARY_PATH.


pkuyym (approved):
LGTM

PR #214, opened 2017-08-21
Add GRU support for DS2 model.

### LINKED ISSUE #213: Add GRU option (instead of only simple RNN) for DS2 model.

xinghai-sun:
Resolve #213

1) Add GRU support for DS2 model.
2) Replace all BRelu activator with Relu
3) Set default GRU size to 1280.
4) Set default striding step of first conv layer to 2 (from 3).

### DIFF
--- a/deep_speech_2/train.py
+++ b/deep_speech_2/train.py
@@ -37,9 +37,14 @@ parser.add_argument(
     help="RNN layer number. (default: %(default)s)")
 parser.add_argument(
     "--rnn_layer_size",
-    default=512,
+    default=1280,
     type=int,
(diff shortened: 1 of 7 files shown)


kuke (commented):
LGTM

deep_speech_2/train.py
```diff
@@ -37,10 +37,15 @@
     help="RNN layer number. (default: %(default)s)")
 parser.add_argument(
     "--rnn_layer_size",
-    default=512,
+    default=1280,
```
pkuyym, line 40: Mark here, replace by actual size later.

pkuyym (approved):
LGTM

PR #224, opened 2017-09-04
Revert back to support input-hidden weights sharing between bi-directional RNNs.

xinghai-sun:
1. Add options to enable and disable RNN weights sharing.
2. Set rnn_layer_size to 2048 by default.
3. Revert back the striding steps of 1st conv layer from 2 to 3.
4. Revert Back to BRelu.

Above follows DS2 papers.


pkuyym (approved):
LGTM

PR #236, opened 2017-09-07
fix missing import

tensor-tang:
fix issue when infer ds2:

> 'ctc_beam_search_decoder_batch' is not defined


xinghai-sun (approved):
LGTM. Thanks a lot!

PR #252, opened 2017-09-14
update README.md and train.py

ranqiu92:


lcy-seso (approved):
Thanks for the refine.

PR #270, opened 2017-09-17
fix bugs for model.py and demo_server.py.

pkuyym:


deep_speech_2/model_utils/model.py
```diff
@@ -79,7 +80,7 @@ def train(self,
         """
         # prepare model output directory
         if not os.path.exists(output_model_dir):
-            os.mkdir(output_model_dir)
+            mkpath(output_model_dir)
```
xinghai-sun, line 83: Use `os.makedirs(output_model_dir)` ?

xinghai-sun (approved):
LGTM

deep_speech_2/model_utils/model.py
pkuyym: https://stackoverflow.com/questions/600268/mkdir-p-functionality-in-python

PR #307, opened 2017-09-21
fix a deep speech 2 speed bug

### LINKED ISSUE #306: Speed bug in DeepSpeech2 data module
[This line](https://github.com/PaddlePaddle/models/blob/88edc4c9f7d232a8a574783dc3b01afc697d8f06/deep_speech_2/data_utils/data.py#L330)
```
batch_manifest = list(sum(batch_manifest, ()))
```  
is O(N^2) complexity. It took **very very** significant time during training.
I gave a small demo [gist](https://gist.github.com/lispc/682910e2c445b5fb33f59dd59f11dfb4) code for verifying this.
And the [stackoverflow link](https://stackoverflow.com/questions/716477/join-list-of-lists-in-python#comment8701078_716489) also discussed this.   

I will make a pull request to fix it.

lispc:
fix https://github.com/PaddlePaddle/models/issues/306

### DIFF
--- a/deep_speech_2/data_utils/data.py
+++ b/deep_speech_2/data_utils/data.py
@@ -327,7 +327,7 @@ class DataGenerator(object):
         shift_len = self._rng.randint(0, batch_size - 1)
         batch_manifest = zip(*[iter(manifest[shift_len:])] * batch_size)
         self._rng.shuffle(batch_manifest)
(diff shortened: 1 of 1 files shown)


lcy-seso (approved):
Thank you for this improvement, a small line, but helps a lot !

PR #355, opened 2017-10-07
Add multi-process version of xmap_reader to take place of multi-thread version for a significant acceleration and add seqbin data parser for internal 1w data training.

### LINKED ISSUE #354: [Deep Speech 2] Slow training with internal English dataset
Currently, training with our internal English dataset (.seqbin) is unexpectedly slow. Only 15% GPU utilization. The normal speed should be larger than 70% GPU utilization.

By profiling, we found the reason is:

Some audio data in this dataset requires a resampling (from 8000 to 16000 sample-rates) before spectrogram feature extraction. Such a resampling is CPU intensive. However, `paddle.reader.xmap_readers` is multi-threading, which can in fact use only single CPU core due to GIL (refer to [Link](https://stackoverflow.com/questions/8774989/python-multithreading-too-slow-multiprocess)).

For LibriSpeech dataset, this problem was not revealed since no CPU intensive resampling is needed for LibriSpeech data.

In a word, we need a multiprocessing version of `xmap_reader`.

xinghai-sun:
Resolved #354 

Done:
1. Add multi-process version of `xmap_reader_mp` to take place of multi-thread version of `paddle.reader.xmap_reader`.
2. Add seqbin data parser to adapt training with internal 1w English dataset.

This speeds up the training with the internal 1w dataset by more than 3X, and now the GPU utilization rises back to 81% (from 15%).


deep_speech_2/data_utils/audio.py
```diff
@@ -115,6 +117,46 @@ def slice_from_file(cls, file, start=None, end=None):
         return cls(data, sample_rate)
 
     @classmethod
+    def from_sequence_file(cls, filepath):
+        """Create audio segment from sequence file.
```
pkuyym: Please add some simple comments about sequence file.

deep_speech_2/data_utils/utility.py
```diff
@@ -61,3 +64,100 @@ def unpack(filepath, target_dir, rm_tar=False):
...
+            w.daemon = True
+            w.start()
+
+        # get results
+        sample = out_queue.get()
+        while not isinstance(sample, XmapEndSignal):
+            yield sample
+            sample = out_queue.get()
+        finish = 1
+        while finish < process_num:
```
pkuyym: I'm not sure why we need two while loop here? The only exit condition is ```finish >= process_num```.

pkuyym (commented):
Great! Almost LGTM.

deep_speech_2/data_utils/utility.py
xinghai-sun: Yes. We can remove the first loop and set `finish = 0` for Line 155, without any difference.

deep_speech_2/data_utils/audio.py
xinghai-sun: Done.

pkuyym (approved):
LGTM

PR #392, opened 2017-10-20
Fix model saving bug

pengli09:
Models should be saved by `trainer.save_parameter_to_tar(f)` instead of `parameters.to_tar(f)`. Otherwise, model average will not take effect.


lcy-seso (approved):
LGTM.

PR #533, opened 2017-12-11
Fix resnet_cifar10 in image_classification.

### LINKED ISSUE #524: The ResNet config for cifar10 in image_classification/resnet.py is not right.
https://github.com/PaddlePaddle/models/blob/3a6d980dae529d74e1dd76be26f633b383969e92/image_classification/resnet.py#L88-L90

It should be as follows, remove the input channel number. Since the `layer_warp` interface has changed. 

```python
    res1 = layer_warp(basicblock, conv1, 16, n, 1)
    res2 = layer_warp(basicblock, res1, 32, n, 2)
    res3 = layer_warp(basicblock, res2, 64, n, 2)
```

peterzhang2029:
resolve https://github.com/PaddlePaddle/models/issues/524


qingqing01 (approved):
LGTM.

PR #534, opened 2017-12-11
Fix the output of cnn in DSSM.

### LINKED ISSUE #526: bug in cnn-rank-dssm
in line 154 of models/dssm/network_conf.py:
return paddle.layer.concat(input=[conv_3, conv_4])   
rather than:
return conv_3, conv_4

peterzhang2029:
resolve https://github.com/PaddlePaddle/models/issues/526

### DIFF
--- a/dssm/network_conf.py
+++ b/dssm/network_conf.py
@@ -151,7 +151,7 @@ class DSSM(object):
         logger.info("create a sequence_conv_pool which context width is 4")
         conv_4 = create_conv(4, self.dnn_dims[1], "cnn")
 
-        return conv_3, conv_4
+        return paddle.layer.concat(input=[conv_3, conv_4])
 
     def create_dnn(self, sent_vec, prefix):
         # if more than three layers, than a fc layer will be added.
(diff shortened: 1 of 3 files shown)


dssm/network_conf.py
```diff
@@ -151,7 +151,7 @@ def create_conv(context_len, hidden_size, prefix):
         logger.info("create a sequence_conv_pool which context width is 4")
         conv_4 = create_conv(4, self.dnn_dims[1], "cnn")
 
-        return conv_3, conv_4
+        return paddle.layer.concat(input=[conv_3, conv_4])
```
lcy-seso: - Could you please also help to fix the logging information in line 151:
    ```python
    logger.info("create a sequence_conv_pool whose context width is 4.")
    ```
- Could you please confirm with @ChunweiYan , is `concat` or `addto` needed here as his original design.

Superjomn (approved):
LGTM

dssm/network_conf.py
peterzhang2029: Done

PR #539, opened 2017-12-12
Implement Inception-v4 for image classification

### LINKED ISSUE #538: Implement Inception-v4 network for image classification
https://arxiv.org/pdf/1602.07261.pdf

will-am:
resolve #538


image_classification/inception_v4.py
```diff
@@ -0,0 +1,526 @@
...
+        input=b2_conv2,
+        num_channels=320,
+        num_filters=320,
+        filter_size=3,
+        stride=2,
+        padding=1)
+    return paddle.layer.concat(input=[b0_pool0, b1_conv1, b2_conv3])
+
+
+def inception_v4(input, class_dim):
```
guoshengCS, line 501: I have tried this configuration and it can work without errors.  But there are some points I am not sure whether appropriate. It seems that the original inception-v4 model uses input with size `3 * 299 * 299`, while the size `3 * 224 * 224` here will lead to some diffs like padding sizes and not be rigorously same. Additionally, I just compare the configuration with the TF implementation and I see some blocks like `Incpetion_A` and `Recution_A` have different structures with the TF implementation. I am not sure if it is appropriate.

image_classification/inception_v4.py
```diff
@@ -0,0 +1,526 @@
...
+        return paddle.layer.concat(input=[conv0, pool0])
+
+    conv3 = block0(conv2)
+    conv4 = block1(conv3)
+    conv5 = block2(conv4)
+    return conv5
+
+
+def Inception_A(input, depth):
+    b0_pool0 = paddle.layer.img_pool(
```
guoshengCS, line 147: I just check with the TF implementation, and it seems that there are some diffs in [`inception_a`](https://github.com/tensorflow/models/blob/f87a58cd96d45de73c9a8330a06b2ab56749a7fa/research/slim/nets/inception_v4.py#L34), like `branch0` excludes `pooling` and `branch3` includes `pooling`.

image_classification/inception_v4.py
```diff
@@ -0,0 +1,526 @@
...
+        input=b3_conv1,
+        num_channels=96,
+        num_filters=96,
+        filter_size=3,
+        stride=1,
+        padding=1)
+    return paddle.layer.concat(input=[b0_conv0, b1_conv0, b2_conv1, b3_conv2])
+
+
+def Inception_B(input, depth):
```
guoshengCS, line 214: Similar to `Inception_A`,  the network structure here has some diffs in with the TF implementation of `Inception_B`, like `branch0` excludes `pooling` and `branch3` includes `pooling`.

image_classification/inception_v4.py
```diff
@@ -0,0 +1,526 @@
...
+        input=b3_conv3,
+        num_channels=224,
+        num_filters=256,
+        filter_size=(7, 1),
+        stride=1,
+        padding=(3, 0))
+    return paddle.layer.concat(input=[b0_conv0, b1_conv0, b2_conv2, b3_conv4])
+
+
+def Inception_C(input, depth):
```
guoshengCS, line 306: Similar to `Inception_A`,  the network structure here has some diffs in with the TF implementation of `Inception_C`, like `branch0` excludes `pooling` and `branch3` includes `pooling`.

image_classification/inception_v4.py
```diff
@@ -0,0 +1,526 @@
...
+        num_channels=512,
+        num_filters=256,
+        filter_size=(1, 3),
+        stride=1,
+        padding=(0, 1))
+    return paddle.layer.concat(
+        input=[b0_conv0, b1_conv0, b2_conv1, b2_conv2, b3_conv3, b3_conv4])
+
+
+def Reduction_A(input):
```
guoshengCS, line 399: The network structure here has some diffs in with the TF implementation of `Reduction_A `.

image_classification/inception_v4.py
```diff
@@ -0,0 +1,526 @@
...
+        input=b2_conv1,
+        num_channels=224,
+        num_filters=256,
+        filter_size=3,
+        stride=2,
+        padding=1)
+    return paddle.layer.concat(input=[b0_pool0, b1_conv0, b2_conv2])
+
+
+def Reduction_B(input):
```
guoshengCS, line 442: The network structure here has some diffs in with the TF implementation of `Reduction_B `.

image_classification/inception_v4.py
will-am: Thank you very much for your reviews. The original inception-v4  model proposed in this paper (https://arxiv.org/pdf/1602.07261.pdf) indeed uses `3 * 299 * 299` as the input image size. However, the paddle examples of image classification utilize the flower dataset as training images and their size is `3 * 224 * 224`. Thus, I changed the input size of inception-v4. But the structure of the model is all the same as the original paper.

image_classification/inception_v4.py
will-am: I checked the TF implementation of inception-v4 (https://github.com/tensorflow/models/blob/master/research/slim/nets/inception_v4.py). In fact, this structure is the same as that in TF because the order of branches do not affect the structure. I add the branch as the Figure 4 in the original paper (https://arxiv.org/pdf/1602.07261.pdf) from left to right. However, the TF implementation does not follow the order as that in the figure. But the order of branches does not matter. The structure is in fact the same.

image_classification/inception_v4.py
will-am: Same as above.

image_classification/inception_v4.py
will-am: Same as above

image_classification/inception_v4.py
will-am: Same as above.

image_classification/inception_v4.py
will-am: Same as above.

image_classification/inception_v4.py
guoshengCS: I get. Sorry for this and the late reply.

image_classification/inception_v4.py
guoshengCS: Please add contents in README to illustrate the diff.

will-am: Add the note in README:

```
Inception-v4模型可以通过下面的代码获取, 本例中使用的模型输入大小为`3 * 224 * 224` (原文献中使用的输入大小为`3 * 299 * 299`)
```

guoshengCS (approved):
LGTM

PR #577, opened 2018-01-17
Add fluid version of SE-ResNeXt

will-am:


qingqing01 (approved):
I approve this PR, since we will train the model based on this config on ImageNet 2012. If any problem, it can be fixed later.

PR #580, opened 2018-01-19
use the same version yapf with paddle main repo.

### LINKED ISSUE #581: use the same version yapf with Paddle main repo.
Currently, the Paddle/models repository uses a different version yapf with Paddle main repository. It is quite inconvenient if one contributes codes to both repositories in pre-commit hook.

lcy-seso:
fixes https://github.com/PaddlePaddle/models/issues/581


luotao1 (approved):
LGTM

PR #597, opened 2018-01-24
 Add targeted attack methods.

guangzhuwu:
1. Add an new class "Adversary", which defines a class that contains the original object, the target and the adversarial example.
2. Modify _apply method of IteratorGradientSignAttack and GradientSignAttack to support targeted attack.


fluid/adversarial/mnist_tutorial_fgsm.py
```diff
@@ -76,10 +76,11 @@ def main():
     att = GradientSignAttack(m)
     for data in train_reader():
         # fgsm attack
-        adv_img = att(data)
-        plt.imshow(n[0][0], cmap='Greys_r')
-        plt.show()
-        #np.save('adv_img', adv_img)
+        adversary = att(Adversary(data))
```
gx-wind: The type of data is [(np.numpy, int)]，data[0][0] represents the image and data[0][1] represent the label, so it should be like this?
```python
adversary = att(Adversary(data[0][0], data[0][1]))
```

guangzhuwu: Yep!

fluid/adversarial/advbox/attacks/gradientsign.py
```diff
@@ -2,37 +2,58 @@
...
+                                             adversary.target_label)])
+            gradient_sign = -np.sign(gradient) * (max_ - min_)
+        else:
+            gradient = self.model.gradient([(adversary.original,
+                                             adversary.original_label)])
+            gradient_sign = np.sign(gradient) * (max_ - min_)
         for epsilon in epsilons:
-            adv_img = image_label[0][0].reshape(
-                gradient_sign.shape) + epsilon * gradient_sign
+            adv_img = adversary.original + epsilon * gradient_sign
```
gx-wind: The shape of adversary.original and gradient_sign may be not same.For example, in mnist dataset, the shape of adversary.original is (784,), and the shape of gradient_sign is (1, 1, 28, 28), then it will throw an error:
```sh
ValueError: operands could not be broadcast together with shapes (784,) (1,1,28,28)
```

guangzhuwu:
The shape bug is fixed.

gx-wind (approved):
LGTM!

PR #652, opened 2018-02-07
Make gradient attack method support norm (L1/L2/L∞, etc.).

guangzhuwu:
Make gradient attack method support norm (L1/L2/L∞, etc.).
Add some method's short name, such as FGSM/FGSMT/BIM/ILCM, etc.
Add lbfgs attack methods


fluid/adversarial/advbox/attacks/gradient_method.py
```diff
@@ -0,0 +1,164 @@
...
+        super(GradientMethodAttack, self).__init__(model)
+        self.support_targeted = support_targeted
+
+    def _apply(self, adversary, norm_ord=np.inf, epsilons=0.01, steps=100):
+        """
+        Apply the gradient attack method.
+        :param adversary(Adversary):
+            The Adversary object.
+        :param norm_ord(int):
+            Order of the norm, such as np.inf, 0, 2, etc.
```
gx-wind: L0 distance measures the number of coordinates i such that xi != xi'.Thus, the L0 distance corresponds to the number of pixels that have been altered in an image. So the norm here means L1, L2 or L∞ like the cleverhans.

guangzhuwu: Yep! Fixed.

gx-wind:
LGTM!

gx-wind (approved):

PR #655, opened 2018-02-07
adversarial example attack method -- jsma

lyzsea:
add a attack method to generate adversarial example.
Jacobian-based Saliency Map Approach (JSMA)


guangzhuwu:
if item != 0:
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

guangzhuwu:
The mask could be a 3 dim array.   "if not any(mask.flatten())"  could be helpful.

guangzhuwu (approved):
LGFM

PR #741, opened 2018-03-19
Add program caching

panyx0718:
Reduce step time from 0.2 to 0.15

transform


guoshengCS (approved):
LGTM. Thanks.  One question about `program_cache` is should it be better that the dict `program_cache` belongs to `program` rather than `executor`.

panyx0718:
Maybe. Or it should belong to something else.

PR #767, opened 2018-03-26
Ssd coco reader

xingyuanbu:
Add coco dataset reader in which coco json image data & annotations were read


fluid/object_detection/train.py
```diff
@@ -131,17 +131,18 @@ def test(pass_id):
...
         mean_value=[127.5, 127.5, 127.5])
     train(args,
-          train_file_list='./data/trainval.txt',
-          val_file_list='./data/test.txt',
-          data_args=data_args,
-          learning_rate=0.001,
-          batch_size=32,
-          num_passes=300)
+        train_file_list='./data/coco/annotations/instances_train2014.json',
+        val_file_list='./data/coco/annotations/instances_val2014.json',
```
qingqing01: Do not specify the COCO data path here.  Should allow the user to specify data by the input arguments.

fluid/object_detection/train.py
```diff
@@ -1,6 +1,6 @@
 import paddle.v2 as paddle
 import paddle.fluid as fluid
-import reader
+import coco_reader as reader
```
qingqing01: It seems coco_reader does not exist.

qingqing01 (approved):

PR #795, opened 2018-03-30
Replace cross_entropy with softmax_with_cross_entropy in Transformer

guoshengCS:
Replace cross_entropy with softmax_with_cross_entropy in Transformer.


pkuyym (approved):
LGTM

PR #809, opened 2018-04-03
Refine the inference to output special tokens optionally in Transformer

guoshengCS:
Refine the inference to output special tokens optionally in Transformer.


pkuyym (approved):
LGTM

PR #816, opened 2018-04-08
Refine SE-ResNeXt model and use ParallelExecutor.

qingqing01:


fluid/image_classification/train.py
```diff
@@ -0,0 +1,310 @@
...
+    else:
+        bd = lr_strategy["bd"]
+        lr = lr_strategy["lr"]
+        optimizer = fluid.optimizer.Momentum(
+            learning_rate=fluid.layers.piecewise_decay(
+                boundaries=bd, values=lr),
+            momentum=0.9,
+            regularization=fluid.regularizer.L2Decay(1e-4))
+
+    opts = optimizer.minimize(avg_cost)
```
chengduoZH: `opts = optimizer.minimize(avg_cost)` should be placed in line 84. Because `minimize` will add  the backward ops.

fluid/image_classification/train.py
```diff
@@ -0,0 +1,310 @@
...
+                       acc1 {3}, acc5 {4} time {5}"
+                                                   .format(pass_id, \
+                       batch_id, loss[0], acc1[0], acc5[0], \
+                       "%2.2f sec" % period))
+                sys.stdout.flush()
+
+        train_loss = np.array(train_info[0]).mean()
+        train_acc1 = np.array(train_info[1]).mean()
+        train_acc5 = np.array(train_info[2]).mean()
+        for data in test_reader():
```
chengduoZH, line 122: Testing phase can be written as a function. Just advise, this is my own habit.

fluid/image_classification/train.py
```diff
@@ -0,0 +1,310 @@
...
+        bd = lr_strategy["bd"]
+        lr = lr_strategy["lr"]
+        optimizer = fluid.optimizer.Momentum(
+            learning_rate=fluid.layers.piecewise_decay(
+                boundaries=bd, values=lr),
+            momentum=0.9,
+            regularization=fluid.regularizer.L2Decay(1e-4))
+
+    opts = optimizer.minimize(avg_cost)
+    fluid.memory_optimize(fluid.default_main_program())
```
chengduoZH: `test_program`'s memory also need to be optimized.

fluid/image_classification/train.py
```diff
@@ -0,0 +1,310 @@
...
+
+    inference_program = fluid.default_main_program().clone()
```
chengduoZH: inference_program should only contain the forward ops. So line 80 should be placed in line62.

qingqing01: Done. But the original code is also right.

fluid/image_classification/train.py
qingqing01: Done. But the `memory_optimize` has never been tested in ParallelExecutor.

fluid/image_classification/train.py
qingqing01: Done.

fluid/image_classification/train.py
```diff
@@ -0,0 +1,311 @@
...
+
+    for pass_id in range(num_passes):
+        train_info = [[], [], []]
+        test_info = [[], [], []]
+        for batch_id, data in enumerate(train_reader()):
+            t1 = time.time()
+            loss, acc1, acc5 = exe.run(
+                fluid.default_main_program(),
+                feed=feeder.feed(data),
+                fetch_list=[avg_cost, acc_top1, acc_top5])
```
chengduoZH, line 105: fetching operation is time-consuming, if possible, we can fetch data every 10 iterations.

chengduoZH (approved):

PR #819, opened 2018-04-08
Avoid predicting <pad> by restricting the size of  fc_layer in Transformer

guoshengCS:
Avoid predicting <pad> by restricting the size of  fc_layer in Transformer.


pkuyym (approved):
LGTM.

PR #838, opened 2018-04-12
rewrite the text-classification model

gmcather:
separate infer.py from train.py, add readme


jshower:
先通过check

jshower (commented):
请参照 https://www.python.org/dev/peps/pep-0008/#imports 调整下import的顺序并合适的加入空行。

jshower (approved):
Thanks.

PR #841, opened 2018-04-13
Add MomentumIteratorAttack to Advbox

buaawht:
* Add MomentumIteratorAttack to Advbox based on [Boosting Adversarial Attacks with Momentum](https://arxiv.org/abs/1710.06081).


gx-wind (approved):
LGTM!

PR #850, opened 2018-04-15
Create mt_with_external_memory/README_en.md

### LINKED ISSUE #713: Translation Plan-带外部记忆机制的神经机器翻译-汉译英
## 任务背景
针对v2版本，目前我们已有很多中文使用文档，但仍有大量的汉译英工作及少量英译汉工作没有完成，我们希望将中英文版本补充完全，提升用户的阅读体验，也有益于国外用户在使用时查阅文档。
此外，我们的模型库Models，仍有大部分英译汉和汉译英的工作，亟待完成。

## 任务描述
1. 需要做什么：
把“带外部记忆机制的神经机器翻译”**汉译英**

2. 地址
- 中文地址： [Github地址](https://github.com/PaddlePaddle/models/blob/develop/mt_with_external_memory/README.md)，[网页地址](http://www.paddlepaddle.org/docs/develop/models/mt_with_external_memory/README.html)
- 需要翻译的英文地址，暂无，可在中文版所在github目录下创建md文档

## 如何认领
我们欢迎大家积极认领，内部小伙伴请通过self assign的方式认领，外部小伙伴可通过在本issue下方comment认领。

## 注意事项
上交初稿的截止日期是2018年4月15日
上交后，PaddlePaddle的RD会为大家review
请于review后的5个工作日之内完成修改

## 相关信息
所有需要翻译的文档：
V2使用文档：
https://shimo.im/sheet/PqK407VYm80Ju3dw
Models:
https://shimo.im/sheet/6XTZnRQEUS0uC1Ut

ASY246:
fix https://github.com/PaddlePaddle/models/issues/713


shanyi15:
您好，请将README.eng.md改为README_en.md，谢谢

shanyi15:
@xinghai-sun Please review this translation at your free time, especially technical terms
Thank you so much!!

mt_with_external_memory/README.eng.md
```diff
@@ -10,9 +10,9 @@ Such models can not only be used in translation, but also other tasks which need
```
skylarch: line 1: Running sample code in this directory requires PaddelPaddle v0.11.0 and later. If the PaddlePaddle on your device is lower than this version, please follow the instructions in [installation document][1] and make an update.

It introduces a differentiable memory network as an additional memory unit, expands the capacity or bandwidth of the working memory within the neural translation model, assists in the temporary access of information in tasks such as translation, and improves model performance.

Such models can  be used in not only translation but also other tasks that...As an important part...improve the performances of

mt_with_external_memory/README.eng.md
```diff
@@ -10,9 +10,9 @@ Such models can not only be used in translation, but also other tasks which need
 
 The External Memory Mechanism is this article, is mainly refer to **Neural Turing Machine** \[[1](#reference)\]. It is worth mentioning that Neural Turing Machines is only one of the attempts in the area of simulating Memory Mechanism by neural network. Memory Mechanism has long been widely studied, and recently in the background of deep learning a series of valuable work is developed, like Memory Networks, Differentiable Neural Computer(DNC), etc. In this article only Neural Turing Machine mechanism is discussed and implemented.
```
skylarch: In this article, the external memory mechanism refers to the **Neural Turing Machine** \[[1](#reference)\]. method (which will be explained in detail later)
...attempts of neural network to simulate memory mechanism...
In the context of deep learning in recent years, a series of valuable work has emerged, such as Memory Networks and Differentiable Neural Computers (DNC).
In this article, only...

mt_with_external_memory/README.eng.md
```diff
@@ -10,9 +10,9 @@ Such models can not only be used in translation, but also other tasks which need
...
 
-The implementation of this article mainly refers to \[[2](#reference)\], and assumes readers have fully read and understood [Machine Translation chapter](https://github.com/PaddlePaddle/book/tree/develop/08.machine_translation) in PaddlePaddle Book.  
+The implementation of this article mainly refers to \[[2](#reference)\], and assumes readers have fully read and understood [Machine Translation chapter](https://github.com/PaddlePaddle/book/tree/develop/08.machine_translation) in PaddlePaddle Book.
```
skylarch: refers to reference and assumes that the readers have...

mt_with_external_memory/README.eng.md
```diff
@@ -36,15 +36,15 @@ Fig 1. state vector in LSTM as memory path
```
skylarch: line 19: With memory, cognition can be coordinated in time, making complex cognition (deducing, planning, and other mental activities that are different from static perception) possible.

mt_with_external_memory/README.eng.md
```diff
@@ -125,12 +125,12 @@ The structure of decoder, explained as follows：
...
-- `model.py`: Model configuration function including bidirectional GRU encoder(`bidirectional_gru_encoder`), memory enhanced decoder(`memory_enhanced_decoder`), memory enhanced seq2seq(`memory_enhanced_seq2seq`). 
+- `model.py`: Model configuration function including bidirectional GRU encoder(`bidirectional_gru_encoder`), memory enhanced decoder(`memory_enhanced_decoder`), memory enhanced seq2seq(`memory_enhanced_seq2seq`).
 - `data_utils.py`: Data processing functions.
 - `train.py`: Train model.
 - `infer.py`: Translation of some samples(model inference).
 
-### `ExternalMemory` 
+### `ExternalMemory`
 
 `ExternalMemory` implements a generic, simplified version of **Neural Turing Machine**. Compared with full version of Neural Turing Machine, this class only implements Content Addressing interpolation rather than Convolutional Shift(Sharpening). Readers can complement it as a complete Neural Turing Machine by themselves.
```
skylarch: Static Memory

Any machine learning model inherently has static memory ability to some degree: whether it is a parametric model (memory as its parameter) or a non-parametric model (memory as its sample); whether it is a traditional SVM (memory as its support ) or neural Network model (memory as its network connection weight). However, “memory” here mostly refers to static memory, meaning that memory is solidified after the model training is finished. When inferred, the model is statically consistent and does not possess the ability to memorize additional information across time steps.

Dynamic Memory 1 --- Hidden State Vectors in RNNs

When dealing with sequence-cognitive problems (such as natural language processing, sequence decisions, etc.), since information processing at each time step depends on the information of other time steps, we often need to maintain a lasting information path at different time steps. Recurrent Neural Networks (RNNs) with a hidden state vector h (or state c in LSTM) have such dynamic memory capabilities. At each time step, the model can obtain the “memory” information of past time steps from h or c, and can continue to contribute new information to update the memory. When the model is inferred, different samples are dynamic and have completely different sets of memory information (h or c).

Despite that the intuitive explanation of cell state c in LSTM meantioned above has a lot of imprecision: for example, from the optimization point of view, the introduction of c or the introduction of the linear Leaky structure in GRU is to make the spectral distribution of the single step gradient Jacobian matrix closer to the unit matrix in the gradient calculation. As a result, this operation alleviates the long-term gradient attenuation problem and reduces the optimization difficulty. But this does not prevent us from intuitively understanding the function of c as increasing the “linear path” to make the “memory path” smoother. As shown in Figure 1, the cell state vector c in the LSTM (cited from this article) can be considered as a "linear memory path" for information to last.


Figure 1. Cell state vector in LSTM as a "memory path"

Dynamic Memory 2 --- Attention Mechanism in Seq2Seq

However, the information bandwidth of a single vector h or c described in the previous section is limited. In the sequence-to-sequence generation model, such a bandwidth bottleneck is manifested in the process of transferring information from an encoder to a decoder: relying only on a finite-length state vector to encode the entire variable-length source statement might incur a large information lost. (仅仅依赖一个有限长度的状态向量来编码整个变长的源语句，有着较大的潜在信息丢失。)

[3] Attention mechanism can overcome the above difficulties. When decoding, the decoder no longer solely relies on the information of the unique sentence-level encoding vector from the encoder, but rather relies on the memory information of a vector group: each vector in the vector group is a coding vector (eg. ht) of a character (Token) of the encoder. Attention are dynamically allocated through a set of learnable attentional weights, and information is read in a linearly weighted manner for symbol generation at different time steps of the sequence (see the chapter about Machine Translation in PaddlePaddle Book). This distribution of attention intensity can be seen as content-based addressing (please refer to the description of addressing in Neural Turing Machine [1]), that is, we determine the reading intensity at different positions of the source statement according to the content, making the distribution of our attention "soft align" with the source statement.

Comparing to single state vectors, a vector group contains more accurate information. For example, a vector group can be considered as an unbounded external memory, which effectively broadens the memory information bandwidth. "Unbounded" refers to the fact that instead of being fixed, the number of vectors in the vector group changes with the number of characters in the source sentence and is not limited. When the encoding of the source sentence is completed, the external storage is initialized to the state vector of each character, and is read and used in the subsequent decoding process.

Dynamic Memory 3 --- Neural Turing Machine (NTM)
Turing Machine or Von Neumann Architecture is the prototype of computer architecture. Operators (such as algebraic calculations), controllers (such as logic branch control) and memory are integrated; together, they form the core operating mechanism of contemporary computers. While most of the common machine learning models ignore explicit dynamic storage, NTMs [1] try to make up for such a potential defect and realise more complex intelligence by using a neural network to simulate a Turing machine that can be differentiated (ie, can learn by gradient descent). 


Figure 2. Comics of Turing machine structure 

Turing machines' storage mechanism is often compared to read and write operations on a tape. The Read Head and Write Head are responsible for reading and writing information on the tape; the movement of the paper bag, the read/write action of the read/write head, and the content are controlled by the controller (see FIG. 2, from here). The length of the tape is usually limited.

NTMs simulate “paper tape” with a matrix M∈n×m, where n is the number of memory vectors (also又称 memory slots) and m is the length of the memory vector. A feedforward neural network or a recurrent neural network is used to simulate the controller and determine the read-and-write intensities at different memory slots, ie, addressing (please refer to paper [1] for details).:

Content-based Addressing: The intensity depends on the content of the memory slot and the actual content of this read-and-write operation;
Location-based Addressing: The intensity depends on the addressing strength of the last addressing operation (eg, offset);
Hybrid addressing: Mix the above addressing methods (eg, linear interpolation);

Based on the addressing situation, the Turing machine writes M or reads information from M for other networks to use. The structural diagram of the NTM is shown in Figure 3 and is cited in [1].


Figure 3. Schematic diagram of the NTM

Comparing with the attention mechanism in the previous section, NTMs have many similarities and differences. Similar points include:

Both use external storage in the form of a matrix (or groups of vectors).
Both use differentiable addressing.

The differences are:

While the attention mechanism initializes the stored contents right when the encoding is complete (caching the contents, not a differentiable writing operation) and only does reading operations in the subsequent decoding process, an NTM can both read and write, making it functioning as memory in the real sense.
NTMs have not only content-based addressing but also location-based addressing, making tasks that require sequential addressing such as sequence replication easier. The attention mechanism, however, only considers content-based addressing that realises soft alignment （中文此处拼写错误）.
NTMs use bounded storage; attention mechanisms use unbounded storage.

Mixing the three ways of memorization to enhance the neural machine translation model

(中文标点错误）Although the attention mechanism is part of the standard configuration of a common sequence-to-sequence model, the external memory of the attention mechanism is only used to store the encoder information. Inside the decoder, the information path still depends on the state vector h or c of the RNN. Naturally, the external storage mechanism of NTMs is used to complement the singl- vector information path inside the decoder.

Thus, we mixed the three dynamic memory mechanisms described above. The original state vector of the RNN and the attention mechanism are preserved; meanwhile, a bounded external memory mechanism based on a simplified version of NTMs was introduced to supplement the decoder's single state vector memory. The implementation of the resulting model is explained in this paper [2]. For detailed implementation differences, see other discussions.

Here is another point worth noting: why not increase the information bandwidth directly by increasing the dimension of h or c?

Increasing the dimension of h or c results in O(n2) complexity on storage and computation (using a state-to-state transfer matrix), whereas the memory expansion cost based on NTMs is O(n) since the unit for addressing is memory slot, and the controller's parameter structure is solely related to the size of the memory slot, m.
Read-and-write memory mechanisms based on single state vector have only a unique read-and-write intensity, which is essentially global. An NTM's mechanism is local, that is, read-and-write operations only happen at part of the memory slots. Locality makes memory access cleaner and less disturbing.

Model Network Structure

The overall network architecture is a simplified version of NTM's [1] external memory module superimposed on a sequence-to-sequence structure with attention mechanisms (ie, RNNsearch [3]).

The encoder uses a standard bidirectional GRU structure (not a stack).
The decoder uses the same structure explained in this paper [2], see Figure 4 (modified from the paper [2]).

Figure 4. Schematic of decoder structure enhanced by external memory

Decoder's structure is explained as follows:

MBt−1 and MBt are bounded external memory matrices. The former is the state of the last time step memory matrix and the latter is the state of the current time step. readB and write are the corresponding read/write heads (including their controllers). Rt is the corresponding readout vector.

MS is an unbounded external memory matrix, and readS is the corresponding read head. The cooperation of the two achieves a traditional attention mechanism. Ct is the corresponding readout vector.

Yt−1 is the word embedded output character of the decoder's previous step. As the input of the current step, yt is the probability distribution of the decoded symbols of the decoder at the current step.

Everything in the dashed box (except the MS) can together be regarded as a bounded external storage module. Without this part, the network structure is basically the same as that of RNNsearch [3] (a small difference: the decoder state used for attention is improved, that is, a hidden layer is superimposed and yt−1 is introduced).

Algorithm Implementation

The algorithm is implemented in the following files:

External_memory.py: The main implementation of the simplified version of NTM in the ExternalMemory class, providing initialization and read-write functions.
Model.py: Related model configuration functions, including bidirectional_gru_encoder, memory_enhanced_decoder with external memory enhancement, and memory_enhanced_seq2seq with external memory enhancement.
Data_utils.py: related data processing helper functions.
Train.py: Model training.
Infer.py: Translation of some sample samples (model inference).

The ExternalMemory Class

The ExternalMemory class implements a generic, simplified version of the NeuroTuring machine. Comparing to the full version of an NTM, this class only implements content addressing rather than convolutional shift. Readers can complement it as a complete NTM.

The structure is as follows:

mt_with_external_memory/README.eng.md
```diff
@@ -0,0 +1,485 @@
...
+#### Differences of implementation with\[[2](#reference)\]
+
+The differences are：
+
+1. The content based addressing formula is different: the original is $a = v^T(WM^B + Us)$, and in this example is  $a = v^T \textrm{tanh} (WM^B + Us) $ to keep the same as the attention mechanism addressing mode in \[[3](#reference)\].
+2. The initialization method of the bounded external storage is different: the original is $M^B = \sigma (W\sum_{i=0}^{i=n}h_i) /n + V$, $V_{i, j}~\in \mathcal{N} (0, 0.1) $, and in this example is $M^B = \sigma (\frac{1}{n}W\sum_{i=0}^{i=n}h_i) + V$.
+3. The addressing logic of external memory mechanism of read-write is different: In original paper they share the same addressing strength, which is equivalent to the Weight Tying regularization. This example does not apply this rule, with reading and writing independently.
+4. The order of external memory in the same time step is different: In original it is first reading and then writing. The example is first writing and then reading. In essence they are equivalent.
+
+## Reference
```
skylarch: Private methods include:

```python
class ExternalMemory(object):
    """External neural memory class.
A simplified Neural Turing Machines (NTM) with only content-based
    addressing (including content addressing and interpolation, but excluding
    convolutional shift and sharpening). It serves as an external differential
    memory bank, with differential write/read head controllers to store
    and read information dynamically as needed. Simple feedforward networks are
    used as the write/read head controllers.
For more details, please refer to
    `Neural Turing Machines <https://arxiv.org/abs/1410.5401>`_.
    """
def __init__(self,
                 name,
                 mem_slot_size,
                 boot_layer,
                 initial_weight,
                 readonly=False,
                 enable_interpolation=True):
        """ Initialization.
:param name: Memory name.
        :type name: basestring
        :param mem_slot_size: Size of memory slot/vector.
        :type mem_slot_size: int
        :param boot_layer: Boot layer for initializing the external memory. The
                           sequence layer has sequence length indicating the number
                           of memory slots, and size as memory slot size.
        :type boot_layer: LayerOutput
        :param initial_weight: Initializer for addressing weights.
        :type initial_weight: LayerOutput
        :param readonly: If true, the memory is read-only, and write function cannot
                         be called. Default is false.
        :type readonly: bool
        :param enable_interpolation: If set true, the read/write addressing weights
                                     will be interpolated with the weights in the
                                     last step, with the affine coefficients being
                                     a learnable gate function.
        :type enable_interpolation: bool
        """
        pass
def _content_addressing(self, key_vector):
        """Get write/read head's addressing weights via content-based addressing.
        """
        pass
def _interpolation(self, head_name, key_vector, addressing_weight):
        """Interpolate between previous and current addressing weights.
        """
        pass
def _get_addressing_weight(self, head_name, key_vector):
        """Get final addressing weights for read/write heads, including content
        addressing and interpolation.
        """
        pass
def write(self, write_key):
        """Write onto the external memory.
        It cannot be called if "readonly" set True.
:param write_key: Key vector for write heads to generate writing
                          content and addressing signals.
        :type write_key: LayerOutput
        """
        pass
def read(self, read_key):
        """Read from the external memory.
:param write_key: Key vector for read head to generate addressing
                          signals.
        :type write_key: LayerOutput
        :return: Content (vector) read from external memory.
        :rtype: LayerOutput
        """
        pass
```

- `_content_addressing`: calculates the addressing intensity of read and write operations through content-based addressing.
- `_interpolation`: updates the current addressing intensity with interpolation (linear weighting of the addressing intensity at current and previous time steps).
- `_get_addressing_weight`: invokes the above two addressing operations to obtain the final addressing intensity of the read and write operations applied to the memory cell.


External interfaces include:

- `__init__`: initializes the class instance.
    - Input parameter `name`: the name of the external memory unit; different instances with the same name will share the same external memory unit.
    - Input parameter `mem_slot_size`: the dimension of a single memory slot (vector).
    - Input parameter `boot_layer`: the layer used for memory slot initialization. It needs to be a sequence type with length of the number of memory slots.
    - Input parameter `initial_weight`: used to initialize the addressing intensity.
    - Input parameter `readonly`: determines whether to open the read-only mode (for example, read-only mode can be used for attention mechanism). If the read-only mode is turned on, then the `write` method cannot be called.
    - Input parameter `enable_interpolation`: determines whether to allow interpolation addressing (for example, when `__init__` is used for attention mechanism, we need to close the interpolation address).
- `write`: the write operation.
    - Input parameter `write_key`: the output of a layer, which contains information used for addressing the write head and generating the actual writing contents.
- `read`: the read operation.
    - Input parameter `read_key`: the output of a layer, which contains information for addressing the read head.
    - Returns: the information read (can be directly used as input to other layers).

Some key implementation logic:

- The "external memory matrix" of Neuro Turing Machine is implemented using `Paddle.layer.memory`. The length of the sequence indicates the number of memory slots, and the `size` of the sequence indicates the size of the memory slot (vector). The sequence relies on an external layer for initialization. The number of memory slots depends on the length of the output sequence of the layer. Therefore, this class can be used to implement not only bounded memory but also unbounded memory (ie, variable number of memory slots).

   ```python
   Self.external_memory = paddle.layer.memory(
       Name=self.name,
       Size=self.mem_slot_size,
       Boot_layer=boot_layer)
   ```
- The addressing logic of the `ExternalMemory` class is implemented via two private methods, `_content_addressing` and `_interpolation`. Read and write operations are implemented using `read` and `write` functions with the addressing operations mentioned above. Unlike in [\[[2](#refs)\] where the read addressing and write addressing share the same adressing intensity,the two addressing in the `ExternalMemory` class proceed independently to make the class more generic. 
- For the sake of simplicity, the controller is not modularized. Instead, it is dispersed in various addressing and read-write functions. The controller includes the write/erase vector generated during the addressing operation and the write operation. The addressing operation is implemented by the two private methods, `_content_addressing` and `_interpolation`. The write/erase vector is generated in the `write` method. The controller mentioned above is a simple feedforward network simulation controller. The reader can try to use a recurrent neural network as controller by stripping off the controller logic and modularizing the controller. 
- The `ExternalMemory` class has a read-only mode. The differential addressing operation can be turned off. This mode can facilitate the implementation of traditional attention mechanism.

- `ExternalMemory` has to be used in conjunction with `paddle.layer.recurrent_group`, specifically in the user-defined `step` function (detailed code is shown in the example), which cannot exist alone.

### `memory_enhanced_seq2seq` and Related Functions

There are three main functions involved:

- The `bidirectional_gru_encoder` function implements a bi-directional single layer GRU (Gated Recurrent Unit) encoder. Two sets of results are returned: one set of character-level encoded vector sequences (both forward and backward) and one set of sentence-level encoded vectors (backward only) for the entire source statement. The former is used to initialize the memory matrix in the attentional mechanism of the decoder, and the latter is used to initialize the state vector of the decoder.

- The `memory_enhanced_decoder` function implements an external memory enhanced GRU decoder. It uses the same `ExternalMemory` class to implement two external memory modules:

    - Unbounded external memory: the traditional attention mechanism. In this module, we turn on the read-only switch, turn off interpolation addressing, and use the first output of the decoder as the initialization of the storage matrix in `ExternalMemory` (`boot_layer`). Therefore, the number of memory slots stored is dynamic and is dependent on the number of characters in the encoder.

        ```python
        Unbounded_memory = ExternalMemory(
            Name="unbounded_memory",
            Mem_slot_size=size * 2,
            Boot_layer=unbounded_memory_init,
            Initial_weight=unbounded_memory_weight_init,
            Readonly=True,
            Enable_interpolation=False)
        ```
    - Bounded external memory: In this module, we turn off the read-only switch and turn on interpolation addressing. After pooling the mean of the decoder's first set of outputs and expanding the result to the specified sequence length, we superimpose random noise (consistent with the value in training and inference phases) to initialize the storage matrix in `ExternalMemory` (`boot_layer` `). The number of memory slots stored is fixed, as in this code snippet:

        ```python
        Bounded_memory = ExternalMemory(
            Name="bounded_memory",
            Mem_slot_size=size,
            Boot_layer=bounded_memory_init,
            Initial_weight=bounded_memory_weight_init,
            Readonly=False,
            Enable_interpolation=True)
        ```

    Note that in our implementation, the attention mechanism (or unbounded external storage) and NTM (or bounded external storage) are implemented as the same `ExternalMemory` class. The former is **read-only** and the latter can **both read and write**. This process helps us unify our understanding of the "attention mechanism" and "memory mechanism" and provides a more concise and consistent implementation version. Attention mechanisms can also be implemented with `paddle.networks.simple_attention`.

- The `memory_enhanced_seq2seq` function defines the entire sequence-to-sequence model with an external memory mechanism, which is the main function of the model definition. It first calls `bidirectional_gru_encoder` to encode the source language and then decode it with `memory_enhanced_decoder`.



In addition, in this implementation, the `write` operation of ʻExternalMemory` happens before the `read` operation to avoid potential limitations on topological connection. See more details in [Issue] (https://github.com/PaddlePaddle/Paddle/ Issues/2061). We can see that they are essentially equivalent.

## Quick Start

### Data Customization

Data is input to the training process through the parameterless iterator() function. Therefore, we need to construct two `reader()` iterators for training data and test data, respectively. The `reader()` function uses `yield` to implement the iterator function (that is, iterates through `for instance in reader()`). For example:

```python
Def reader():
    For instance in data_list:
        Yield instance
```

Each sample returned by `yield` must be a triple containing a list of input characters of the encoder (ie, the source language sequence, which needs to be converted to IDs), a decoder input character list (ie, the target language sequence needing to be converted to IDs. The sequence needs to be shifted a bit to the right), and a decoder output character list (ie, the target language sequence, which needs to be converted to IDs).

The user needs to complete the tokenization of the character and build a dictionary to complete the conversion to IDs.

PaddlePaddle interface [paddle.paddle.wmt14] (https://github.com/PaddlePaddle/Paddle/blob/develop/python/paddle/v2/dataset/wmt14.py) by default provides a preprocessed, small-scale subset of [wmt14 English-French translation data sets] (http://paddlepaddle.bj.bcebos.com/demo/wmt_shrinked_data/wmt14.tgz) (the data set has 193,319 training data, 6003 test data, and a dictionary with length 30000). It also provides two reader creator functions as follows:

When these two functions are called, the corresponding `reader()` function is returned for use by `paddle.traner.SGD.train`. When needing to use other data, we can refer to [paddle.paddle.wmt14](https://github.com/PaddlePaddle/Paddle/blob/develop/python/paddle/v2/dataset/wmt14.py) to construct the corresponding data creator and replace `paddle.dataset.wmt14.train` and `paddle.dataset.wmt14.train` with the corresponding function name.

### Training

Run the following command:

```bash
Python train.py
```
You can also customize some of the parameters, for example:

```bash
Python train.py \
--dict_size 30000 \
--word_vec_dim 512 \
--hidden_size 1024 \
--memory_slot_num 8 \
--use_gpu False \
--trainer_count 1 \
--num_passes 100 \
--batch_size 128 \
--memory_perturb_stddev 0.1
```

The code will run the training script. The training model will be periodically saved to the local `./checkpoints`. To get the meaning of the parameters, run:

```bash
Python train.py --help
```


### Decoding

Run the following command:

```bash
Python infer.py
```
You can also customize some of the parameters, for example:

```bash
Python infer.py \
--dict_size 30000 \
--word_vec_dim 512 \
--hidden_size 1024 \
--memory_slot_num 8 \
--use_gpu False \
--trainer_count 1 \
--memory_perturb_stddev 0.1 \
--infer_data_num 10 \
--model_filepath checkpoints/params.latest.tar.gz \
--beam_size 3
```

You can run the decoding script and generate sample translations. To get the meaning of the parameters, run:

```bash
Python infer.py --help
```


## Other Discussions

Differences between this Method #### and Essay\[[2](#ref)\]

The differences are as follows:

1. The content-based addressing formula is different: the original text is $a = v^T(WM^B + Us)$; in this example, it is $a = v^T \textrm{tanh}(WM^B + Us)$, which is consistent with the attention mechanism addressing method in \[[3](#ref)\].
2. The initialization of bounded external storage is different: in the original text, it is $M^B = \sigma(W\sum_{i=0}^{i=n}h_i)/n + V$, $V_{i,j} ~\in \mathcal{N}(0, 0.1)$. In this example, it is $M^B = \sigma(\frac{1}{n}W\sum_{i=0}^{i=n}h_i) + V$.
3. The addressing logic for read and write of the external memory mechanism is different: the original two share the same addressing intensity, which is equivalent to weight tying. This example does not impose this regularity, and read and write operations use independent addressing.
4. The reading and writing order of external memory within the same time step is different: in the original text, write operations are started after read operations are done; in this example, write operations come before read operations. These two orders are essentially equivalent.

## References

skylarch:

PR #859, opened 2018-04-16
Pr/845

reyoung:


fluid/neural_machine_translation/transformer/train.py
```diff
@@ -170,32 +186,57 @@ def test(exe):
...
+                    ModelHyperParams.eos_idx, ModelHyperParams.eos_idx,
+                    ModelHyperParams.n_head, ModelHyperParams.d_model)
+
+                local_scope = train_exe.executor.local_scope(place_id)
+
+                local_scope.find_var(lr_scheduler.learning_rate.name).get_tensor().set(
+                    lr_scheduler.update_learning_rate(),
+                    fluid.CUDAPlace(place_id))
+
+                for var_name in data_input_dict:
```
panyx0718: can we feed data instead of set?

fluid/neural_machine_translation/transformer/train.py
```diff
@@ -170,32 +186,57 @@ def test(exe):
...
+    dev_count = fluid.core.get_cuda_device_count()
+
     for pos_enc_param_name in pos_enc_param_names:
-        pos_enc_param = fluid.global_scope().find_var(
-            pos_enc_param_name).get_tensor()
-        pos_enc_param.set(
-            position_encoding_init(ModelHyperParams.max_length + 1,
-                                   ModelHyperParams.d_model), place)
+        tensor = position_encoding_init(ModelHyperParams.max_length + 1, ModelHyperParams.d_model)
+        for place_id in xrange(dev_count):
```
panyx0718: Maybe we can have 2 arguments for ParallelExecutor.run(), called feed_dict and data_parallel_feed_dict. data_parallel_feed_dict will split the tensor at the batch dimension while feed_dict will replicate the data to each place.
The logics are implemented in C++ so that it's simpler for the users.

Another question is, if the var being fed will be needed in parameter servers, such as learning rate. will feed_dict works?

PR #868, opened 2018-04-19
add README_cn.md for generate_sequence_by_rnn_lm

GoDownToMissYou:
https://github.com/PaddlePaddle/models/issues/718


generate_sequence_by_rnn_lm/README_cn.md
```diff
@@ -0,0 +1,134 @@
+The program example running in this directory needs to use the PaddlePaddle v0.10.0 version.If your PaddlePaddle installation version is below this requirement, update the PaddlePaddle installation version according to the instructions in the [installation document](http://www.paddlepaddle.org/docs/develop/documentation/zh/build_and_install/pip_install_cn.html).
```
skylarch: Running sample code in this directory requires PaddelPaddle v0.10.0 and later. If the PaddlePaddle on your device is lower than this version, please follow the instructions in [installation document](http://www.paddlepaddle.org/docs/develop/documentation/zh/build_and_install/pip_install_cn.html) and make an update.

generate_sequence_by_rnn_lm/README_cn.md
```diff
@@ -0,0 +1,134 @@
...
+-13.9877    一样 。 他 是 我 的 敌人 <e>
+-14.2741    一样 。 他 是 我 的 <e>
+-14.6250    一样 。 他 是 我 的 朋友 <e>```
+among them:
+	1. The first line 81 is looming like a ghost, like death, separated by \t. There are two columns:
+		+ The first column is the sequence number of the input prefix in the training sample set.
+		+ The second column is the input prefix.
+	2. The second ~beam_size + 1 line is the generated result, also separated into two columns with \t:
+		+ The first column is the logarithmic probability of the generated sequence(log probability)
+		+ The second column is the generated text sequence. The normal generation result will end with the symbol ```<e>```. If it does not end with ```<e>```, it means that the maximum sequence length is exceeded and the forced termination is generated.
```
skylarch: Generating Text using a Recurrent Neural Network Language Model

The language model is a probability distribution model. In a nutshell, it is a model used to calculate the probability of a sentence. Using the model, we can determine which word sequence is more likely to occur. We can also predict which word is most likely to occur after a group of given words. The language model is an important basic model in the field of natural language processing.

Application Scenario

Language models are used in many areas, such as:
Automatic Writing: The language model can generate sentences, paragraphs, and chapters by recursively generating the next word based on the existing words.
QA: The language model can generate the answer based on the question.
Machine Translation: The current mainstream machine translation models are mostly based on the encoder-decoder model, where the decoder is a conditional language model used to generate the target language.
Spell Checking: The language model can calculate the probability for a word sequence to occur. Generally, the probability of the sequence at the misspelling is reduced sharply. It can be used to identify spelling errors and provide a correction candidate set.
Speech tagging, syntactic analysis, speech recognition…

About this Example

This example implements a RNN-based language model and generates text using the language model. The directory structure of this example is as follows:

.
├── data
│ └── train_data_examples.txt # Sample data, refer to sample data format, provide your own data
├── config.py # Configuration file, including data, train, infer related configuration
├── generate.py # Predict task script, ie generate text
├── beam_search.py ​​#Beam search algorithm implementation
├── network_conf.py # The various network structures involved in this example are defined in this file. We can modify the model structure by changing this file.
├── reader.py # Read data interface
├── README.md
├── train.py # Training Mission Script
└── utils.py # Define common functions such as: building a dictionary, loading a dictionary, etc.

RNN Language Model

Introduction

RNN is a sequence model. The basic idea is: At time t, we obtain the characteristic representation of this time by putting the hidden layer output at the previous time t−1 and the word vector at time t into the hidden layer. Using this representation, we can get the predict output of time t and recursively repeat the above operations to obtain the output at future time. RNN has a "memory" function, making it good at using the existing text information and historical knowledge. Theoretically, RNN can achieve "long dependence" (ie, knowledge that has been used long ago), but it has been found to be less than ideal in practical applications. Researches show that variants such as LSTM and GRU can improve the RNN memory unit by introducing a gate mechanism. These improvements make up for the difficulties encountered by traditional RNNs in learning long sequences. This example model uses LSTM or GRU and can be modified by configuration. The following figure is a schematic diagram of the idea of circularity in ​​RNN's language model (with LSTM, GRU, etc.):



Model Implementation
The implementation of the RNN language model in this example is as follows:

Defining model parameters: config.py defines parameter variables for the model.
Defining the model structure: rnn_lm function in network_conf.py defines the structure of the model as follows:
Input layer: the word vector layer where embedding happens. The input word (or character) sequence is mapped into a vector.
The middle layer: implements the RNN layer according to the configuration; the embedding vector sequence obtained in the previous step is used as input.
Output layer: uses softmax to normalize the probability of calculating words.
Loss: defines multiple types of cross entropy as the loss function of the model.

Training model: Function ```main``` in ```train.py``` implements the training of the model. The implementation process is as follows:

Input data preparation: create and save dictionaries, constructd readers for train and test data.
Model initialization : includes the structure and parameters of the model.
Trainer construction: the Adam optimization algorithm is used in the demo.
Defining the callback function: builds the event_handler to track changes in the loss of the training process and saves the parameters of the model at the end of each round of training.
Training: Use the trainer to train the model.

Generating text: generate.py implements text generation. The implementation flow is as follows:

Load the trained models and dictionary files.
Read gen_file file with each line as a sentence prefix. 
Use the column search algorithm (Beam Search) to generate text based on the prefix.
Save the generated text and its prefix to the file gen_result.

Instructions for Use

Here's how to run this example:
* 1. Run `python train.py` to start the train model (use LSTM by default) until training is over.
* 2. Run `python generate.py` to start the text generation process. By default, the input text is `data/train_data_examples.txt`, and the generated text is saved to `data/gen_result.txt`.


**If you need to use your own corpora or custom models, you need to modify the configuration in `config.py`. Details and adaptations are as follows: **


### Corpus Adaptation

* Clean corpora: remove spaces, tabs, garbled text, remove numbers, punctuation marks, special symbols etc. as needed.
* Content format: one sentence per line; each word in each line is separated by a space character.
* Configure the following parameters in `config.py` as required:

    ```python
     Train_file = "data/train_data_examples.txt"
     Test_file = ""

     Vocab_file = "data/word_vocab.txt"
     Model_save_dir = "models"
    ```
    1. `train_file`: specifies the path to the training data, ** requires pre-segmentation**.
    2. `test_file`: specifies the path to the test data. If the training data is not empty, the specified test data will be tested at the end of each `pass` training.
    3. `vocab_file`: specifies the path to the dictionary. If the dictionary file does not exist, the training corpus will construct the dictionary by counting the word frequency.
    4. `model_save_dir`: specifies the path to the model's saved output. If the specified folder does not exist, it will be automatically created.

### Strategies for Building a Dictionary
- When the specified dictionary file does not exist, word frequency statistics will be performed on the training data. The automatic construction of the dictionary `config.py` has the following two parameters related to the construction of the dictionary:

    ```python
    Max_word_num = 51200 - 2
    Cutoff_word_fre = 0
    ```
    1. `max_word_num`: specifies how many words are in the dictionary.
    2. `cutoff_word_fre`: the lowest frequency of words in the dictionary that appear in the training corpus.
- If  we define `max_word_num = 5000` and `cutoff_word_fre = 10`, word frequency statistics will show that there are only 3000 words in the training corpus with frequency higher than 10. These 3000 words will eventually be used to form a dictionary.
- When building a dictionary, two special symbols are automatically added:
    1. `<unk>`: not a dictionary word now
    2. `<e>`: end of sentence

    *Note: It should be noted that, though the larger the dictionary is, the richer the content is, the training process will correspondingly take a longer time. Normally, there can be tens or even hundreds of thousands different Chinese words in corpus, so if `max_word_num` value is too small, then ` <unk> ` will take up too high a proportion; if` max_word_num` value is too great, then it will seriously affect the training speed (also accuracy). Therefore, there is also a “word-by-character” training model, which means that each Chinese character is treated as a single word. There are only several thousands of commonly used Chinese characters, making the size of the dictionary not too large and not losing too much information. However, the same Chinese character can have greatly different meanings in different context, which sometimes leads to unsatisfactory model results. It is recommended to test the models and choose from "word training" and "character training" modes according to the actual situation. *

### Model Adaptation and Training

* Modify the network results of the RNN language model as needed by adjusting the following configuration in `config.py`:

    ```python
    Rnn_type = "lstm" # "gru" or "lstm"
    Emb_dim = 256
    Hidden_size = 256
    Stacked_rnn_num = 2
    ```
    1. `rnn_type`: supports "gru" and "lstm", selects which RNN unit to use.
    2. `emb_dim`: sets the dimension of the word vector.
    3. `did_size`: sets the hidden layer size of the RNN unit.
    4. `stacked_rnn_num`: sets the number of stacked RNN units to form a deeper model.

* Run `python train.py` to train the model. The model will be saved to the directory specified by `model_save_dir`.

### Text Generation on Demand

* Adjust the following variables in `config.py` as needed. Details as follows:

    ```python
    Gen_file = "data/train_data_examples.txt"
    Gen_result = "data/gen_result.txt"
    Max_gen_len = 25 # the max number of words to generate
    Beam_size = 5
    Model_path = "models/rnn_lm_pass_00000.tar.gz"
    ```
    1. `gen_file`: specifies the input data file, each line is a sentence prefix, ** requires pre-segmentation**.
    2. `gen_result`: specifies the output file path, and the result will be written to this file.
    3. `max_gen_len`: specifies the maximum length of each sentence generated. If the model cannot generate `<e>`, the generation process will automatically terminate when the `max_gen_len` term is generated.
    4. `beam_size`: the width of each step of the beam search algorithm.
    5. `model_path`: specifies the path to the trained model.

    Among them, `gen_file` holds the text prefix to be generated, each prefix is in one line, such as:

    ```text
    Looming like a ghost, like a god of death
    ```
    We can write the text prefix to be generated into the file in this format;

* Run `python generate.py` to start the beam search algorithm to generate the text for the input prefix. Here are the results generated by the model:

    ```text
    和中文版一样
    ```
    among them:
    1. The first line `中文版`, has two columns, separated by `\t`:
        - The first column is the sequence number of the input prefix in the training sample set.
        - The second column is the input prefix.
    2. The ~beam_size + 1` line is the generated result, also separated into two columns with `\t`:
        - The first column is the log probability of the generated sequence.
        - The second column is the generated text sequence. Normally, the generation result will end with the symbol `<e>`. If it does not end with `<e>`, then the maximum sequence length is exceeded and a forced termination is generated.

skylarch:

PR #878, opened 2018-04-25
Replace feed_dict by feed

kuke:
To avoid the warning: ```feed_dict is deprecated. Please use feed=```


qingqing01 (approved):
@kuke Thanks!

PR #889, opened 2018-04-28
【Fluid models】implement DQN model

TomorrowIsAnOtherDay:
Hi, I'm zhoubo01 in NLP，this is an implementation of DQN model and I use a classic problem (mountain-car) as its benchmark, it has reproduced the same performance as tensorflow.


fluid/DeepQNetwork/expreplay.py
```diff
@@ -0,0 +1,51 @@
...
+        self.action[pos] = exp.action
+        self.reward[pos] = exp.reward
+        self.isOver[pos] = exp.isOver
+
+    def __len__(self):
+        return self._curr_size
+
+    def sample(self, batch_idx):
+        for i, idx in enumerate(batch_idx):
+            while (idx + 1) % self._curr_size == self._curr_pos:
```
yu239-zz: I understand that you want to avoid using the starting position of the circular buffer as the next state because the next state has not been written in the buffer yet. However, this while loop is confusing and inefficient. Perhaps when you initially sample batch_idx, you can sample from len(exp) - 1 and do an index mapping to avoid sampling the last exp in the buffer.

TomorrowIsAnOtherDay: This while loop was created for  two reasons:
1.the next state has not been written in the buffer yet
2.the next state has been written, but it's a state of another episode(after the buffer was full, and some position would be written new data).

yu239-zz (approved):

PR #899, opened 2018-05-10
hsigmoid translation

### LINKED ISSUE #720: Translation Plan-Hsigmoid加速词向量训练-汉译英
## 任务背景
针对v2版本，目前我们已有很多中文使用文档，但仍有大量的汉译英工作及少量英译汉工作没有完成，我们希望将中英文版本补充完全，提升用户的阅读体验，也有益于国外用户在使用时查阅文档。
此外，我们的模型库Models，仍有大部分英译汉和汉译英的工作，亟待完成。

## 任务描述
1. 需要做什么：
把Hsigmoid加速词向量训练”**汉译英**

2. 地址
- 中文地址： [Github地址](https://github.com/PaddlePaddle/models/blob/develop/hsigmoid/README.md)，[网页地址](http://www.paddlepaddle.org/docs/develop/models/hsigmoid/README.html)
- 需要翻译的英文地址，暂无，可在中文版所在github目录下创建md文档

## 如何认领
我们欢迎大家积极认领，内部小伙伴请通过self assign的方式认领，外部小伙伴可通过在本issue下方comment认领。

## 注意事项
上交初稿的截止日期是2018年4月15日
上交后，PaddlePaddle的RD会为大家review
请于review后的5个工作日之内完成修改

## 相关信息
所有需要翻译的文档：
V2使用文档：
https://shimo.im/sheet/PqK407VYm80Ju3dw
Models:
https://shimo.im/sheet/6XTZnRQEUS0uC1Ut

adrianhust:
first submit 
fix #720

### DIFF
--- a/hsigmoid/READ_EN.md
+++ b/hsigmoid/READ_EN.md
@@ -1,0 +1,174 @@
+PaddlePaddle v0.10.0 is a must to run example scripts in this directory. If your paddle version is older, please refer[Installment doc](http://www.paddlepaddle.org/docs/develop/documentation/zh/build_and_install/pip_install_cn.html) to upgrade paddle
+---
+
+# Hsigmoid accelerated training of word vector
+## Background Introduction
+In natural language processing field, usually, using one-hot vector to present word as traditional method, like word dict ['I', 'you', 'like'], [1,0,0],[0,1,0],[0,0,1] three vectors can be used to present 'I','you' and 'like'. Thought this usage is concise, it's easy to produce dimension explosion; And any vectors of two word is orthogonal, so information in vector is limited. To avoid or alleviate weakness of one-hot presentation, now usually we take word vector instead of one-hot presentation, word vector namely word embedding, that is using low-dimension dense vector to replace high-dimension sparse one-hot vector. There are many methos to training word vector, one of them is neural network model, including CBOW,Skip-gram etc. Esentially, these are classification model, when vocabulary, class is large, traditional softmax is much time-consuming. PaddlePaddle applys Hsigmoid Layer, NCE Layer, to accelerate training.Here we mainly introduce how to use Hsigmoid Layer to accelerate training, issues related to word vector please go to PaddlePaddle Book [Chapter word vector](https://github.com/PaddlePaddle/book/tree/develop/04.word2vec)
+
+
+
+##Hsigmoid Layer
+Hsigmoid Layer refered to paper\[[1](#Reference)\], Hsigmoid is Hierarchical-sigmoid, theory of it is to reduce computing complexity by contructing classification binary tree, every leaf node of binary tree is a class,every non-leaf node is a binary classifier. Supposing we get four classes 0,1,2,3, softmax will calculate four classes score seperately, then normalizate it to propability. When classes are large, it's time-consuming to calculate each class propability, Hsigmoid Layer will construct a balanced binary tree based on quantity of classes, as following:
+
+<p align="center">
+<img src="images/binary_tree.png" width="220" hspace='10'/> <img src="images/path_to_1.png" width="220" hspace='10'/> <br/>
+Figure 1. （a）is balanced binary tree，（b）is path from root node to class 1
+</p>
+each non-leaf node is a two-class classifier(sigmoid) in binary tree, if class is 0, then continue judging classes with left child node, if not with right child node, until reach at leaf node. In this mode, each class is a path, like path code if 0, 1 from root to class 1. In training stage, we calculate loss of classifier respectively  by path in order of real class, then summate all loss to final loss. In predicting stage, model will output propability of each non-leaf classifier, we can get path code by propability, then iterate path code to get final predicting class. Computation cost of traditional softmax is N(N is size of vocabulary dict), Hsigmoid can reduce complexity to log(N), detailed theory refers paper \[[1](#Reference)\]。
+
+## data preparing
+### PTB data
+Here we use Penn Treebank (PTB) dataset（[Tomas Mikolovpreprocessing version](http://www.fit.vutbr.cz/~imikolov/rnnlm/simple-examples.tgz)), including train、valid and test three files. Train is used as training data, valid as testing data. Here training is 5-gram model, that is four previous words in each item predicting fifth word. PaddlePaddle supplys python package for PTB dataset[paddle.dataset.imikolov](https://github.com/PaddlePaddle/Paddle/blob/develop/python/paddle/v2/dataset/imikolov.py), automatically download and preprocess data. Preproessing will add begining sign \<s> and ending sign \<e> to each sentence of dataset, then depending on window size(here is 5), each time slide window right from head to tail and generate a data item. Like "I have a dream that one day" can produce \<s> I have a dream、I have a dream that、have a dream that one、a dream that one day、dream that one day \<e>, PaddlePaddle will convert word into id data as output of preprocessing.
+
+
+### Self-defined data
+User can train model with your own dataset, key point for self-defined dataset is to supply reader interface for data processing, reader need to generate a iterator, iterator is for parsing each item in file, return a python list, for example [1, 2, 3, 4, 5], respectively id of first to fourth word in dict, PaddlePaddle will transform this list `paddle.data_type.inter_value` type as input of data layer, a package example as following:
+
+
+```python
+def reader_creator(filename, word_dict, n):
+    def reader():
+        with open(filename) as f:
+            UNK = word_dict['<unk>']
+            for l in f:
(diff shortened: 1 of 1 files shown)


hsigmoid/READ_EN.md
```diff
@@ -0,0 +1,174 @@
+PaddlePaddle v0.10.0 is a must to run example scripts in this directory. If your paddle version is older, please refer[Installment doc](http://www.paddlepaddle.org/docs/develop/documentation/zh/build_and_install/pip_install_cn.html) to upgrade paddle
```
skylarch, line 1: Running sample code in this directory requires PaddelPaddle v0.10.0 and later. If the PaddlePaddle on your device is lower than this version, please follow the instructions in [installation document](http://www.paddlepaddle.org/docs/develop/documentation/zh/build_and_install/pip_install_cn.html) and make an update.

hsigmoid/READ_EN.md
```diff
@@ -0,0 +1,174 @@
...
+
+<p align="center">
+<img src="images/binary_tree.png" width="220" hspace='10'/> <img src="images/path_to_1.png" width="220" hspace='10'/> <br/>
+Figure 1. （a）is balanced binary tree，（b）is path from root node to class 1
+</p>
+each non-leaf node is a two-class classifier(sigmoid) in binary tree, if class is 0, then continue judging classes with left child node, if not with right child node, until reach at leaf node. In this mode, each class is a path, like path code if 0, 1 from root to class 1. In training stage, we calculate loss of classifier respectively  by path in order of real class, then summate all loss to final loss. In predicting stage, model will output propability of each non-leaf classifier, we can get path code by propability, then iterate path code to get final predicting class. Computation cost of traditional softmax is N(N is size of vocabulary dict), Hsigmoid can reduce complexity to log(N), detailed theory refers paper \[[1](#Reference)\]。
+
```
skylarch, line 18: Hsigmoid Accelerated Word Vector Training

Background 

In the field of natural language processing, traditional practices usually use one-hot vectors to represent words. For example, if the dictionary contains 'I', 'you', and 'like', we can use vectors [1, 0, 0], [0, 1, 0] and [0, 0, 1] to represent 'I', 'you' and 'like' respectively. This representation is relatively simple, but when the vocabulary is large, dimensional explosion can easily occur. Besides, since the vectors of any two words are orthogonal to each other, they contain limited information. In order to avoid or reduce such shortcomings, we use word vectors instead of the one-hot expression. Word vectors are word embedding. That is, a low-dimensional, dense, real vector is used to replace the high-dimensional, sparse, one-hot vector. There are many ways to train word vectors. Neural network models are one of them, including CBOW, Skip-gram, etc. These models are essentially classification models. When the word list and the number of categories are large, softmax will be very time consuming. PaddlePaddle provides Hsigmoid Layer and NCE Layer to accelerate the training process of the model. This article mainly introduces how to use Hsigmoid Layer to speed up training. Please refer to the word vector chapter in PaddlePaddle Book for related content.

Hsigmoid Layer

Hsigmoid Layer is cited from this paper[1]. Hsigmoid refers to Hierarchical-sigmoid. The principle is to reduce the computational complexity by constructing a classification binary tree. Each leaf node in the binary tree represents a category, and each non-leaf node represents a two-category classifier. For example, suppose we have a total of four categories: 0, 1, 2, and 3. Softmax will calculate the scores of the four categories and normalize the results to get a probability. When the number of categories is large, it is very time consuming to calculate the probability of each category. Hsigmoid Layer will construct a balanced binary tree according to the number of categories as follows:

  
Figure 1. (a) Balanced binary tree; (b) Path from root node to category 1

Each non-leaf node in the binary tree is a two-class classifier (sigmoid). If the category is 0, choose the left child node and continue the classification. If not, choose the right child until reaching a leaf node. In this way, each category corresponds to a path. For example, the path from root to category 1 is encoded as "0, 1". In the training stage, we calculate the loss of the corresponding classifier according to the path of each real category, and then synthesize all the losses to obtain the final loss. During the prediction phase, the model outputs the probability of each non-leaf node classifier. We can obtain the path's code based on the probability, and then traverse the path's code to obtain the predicted class. The computational complexity of softmax is N (N is the dictionary size), and Hsigmoid can reduce the complexity to log(N). For detailed theoretical details, refer to this paper [1].

hsigmoid/READ_EN.md
```diff
@@ -0,0 +1,174 @@
...
+                idx = idx * 2 + 1  # left child
+
+        predict_lbl = result - dict_size
+        predict_lbls.append(predict_lbl)
+    return predict_lbls
+```
+Format of input data for predicting is same with training, like `have a dream that one`, program will produce a sequence of propability based on `have a dream that`, then decode propability to generate predicting word, `one` as real word, is easy to evaluate. Input of decoding funtion is prediction propability of one batch sample and vocabulary size, inner loop is to decode each item based on its output propability, decoding method follows left 0 right 1 rule, iterates all paths until arrives at leaf node.
+
+
+## Reference
```
skylarch, line 173: Data Preparation

PTB Data

This article uses the Penn Treebank (PTB) data set (Pre-Tomas Mikolov version), which contains three files: train, valid, and test. We use 'train' as training data and 'valid' as test data. This article trains the 5-gram model, which uses the first 4 words of each piece of data to predict the 5th word. PaddlePaddle provides a python package, paddle.dataset.imikolov, corresponding to the PTB dataset, which automatically performs data download and preprocessing. The preprocessing operations will add a start symbol and an end symbol before and after each sentence in the data set and, depending on the window size (in this article, the size is 5), swipe the window to the right from the beginning to the end to generate a piece of data. For instance, "I have a dream that one day" can generate these pieces of data:  I have a dream, I have a dream that, have a dream that one, a dream that one day, dream that one day. PaddlePaddle will output the ID information of the words.

Custom Data
Users can use their own data set to train the model. The most important feature of a custom data set is to implement the reader interface for data processing. The reader needs to generate an iterator that is responsible for parsing each row of data in the file and returning a python list. For example, suppose that 1, 2, 3, and 4 are the IDs of the first to fourth word in the dictionary. PaddlePaddle will convert the list of IDs into paddle.data_type.inter_value type as the data layer input. A package example is as the following:


Network Structure

This article explains how to obtain the word vector by training the N-gram language model. We use the first four words to predict the current word. The network input is the ID of the word in the dictionary. The query word vector table is used to obtain the word vector. We stitch together the four word vectors and access a fully connected hidden layer. Finally, we get to the Hsigmoid layer. The detailed network structure is shown in Figure 2:


Figure 2. Network configuration structure

code is shown as follows:


It should be noted that in PaddlePaddle, the hsigmoid layer stores the learnable parameters as a matrix of [number of classes - 1 × hidden layer vector width]. When forecasting, you need to replace the hsigmoid layer with a full-join operation and fix it with sigmoid. Prediction has the output as a matrix with width of [batch_size x number of categories - 1] (degenerate into a vector when batch_size = 1). Each dimension of a matrix row vector computes the probability of a right child node whose input vector belongs to an internal node. When the full-joined operation loads the parameter matrix learned by the hsigmoid layer, the parameter matrix needs to be transposed once. The code fragment is as follows:


The paddle.layer.mixed in the above code snippet must be entered as paddle.layer.×_projection in PaddlePaddle. Paddle.layer.mixed sums the results of multiple projection (there can be multiple inputs) as output. Parameter W is transposed by paddle.layer.trans_full_matrix_projection when doing matrix multiplication.

Training Phase

The training is relatively simple; we can run python train.py directly. The first time the program runs, it will check if the user's cache folder contains the imikolov data set. If the data set is not present, it will be downloaded automatically. During the running process, model training information will be printed every 100 iterations, which mainly includes training loss and test loss. Each pass will save the model once.

Prediction Stage

Run the following command line:


The parameters are as follows:
 - model_path: Specifies the path where the trained model is located. This parameter is required.
 - batch_size: Predicts the number of samples in parallel at a time. Optional with default value 1.
- use_gpu: Whether to use GPU for prediction. Optional, the default value is False. - trainer_count : The number of threads predicted to use. Optional, default is 1. Note: The number of threads used for prediction must be greater than the number of samples to be predicted in parallel.
The prediction stage obtains the encoding path based on multiple binary classification probabilities and traverses the path to obtain the final prediction category. The logic is as follows:

The format of input data during the prediction phase is the same as it is in the training phase. For instance, for the phrase "have a dream that one", the program will generate a set of probabilities based on the phrase. By inspecting the probabilities, the program predict the output words. The input of the decoding function is the prediction probability of a batch sample and the size of the vocabulary. The loop inside the decoding function decodes the output probability of each sample. The decoding method follows that if meeting a zero, we go to the left child node and if meeting a one, we go to the right child. We continue to traverse the path until reaching a leaf node.

References

skylarch:

PR #907, opened 2018-05-12
remove light_imdb & tiny_imdb

gmcather:
As light_imdb & tiny_imdb are only used in unittest, it should be removed in the formal code.


luotao1 (approved):
LGTM! Thanks very much!

PR #969, opened 2018-06-07
fix tanh problem

gmcather:


gmcather:
activation function shouldn't be used in fc layer that comes before lstm layer

panyx0718 (approved):

PR #973, opened 2018-06-08
Crnn ctc mkldnn benchmarking

sfraczek:


sfraczek:
Sorry! I didn't notice when and how I have accidentally removed the branch...

PR #981, opened 2018-06-12
Update fluid/DeepQNetwork models with Atari environment

zenghsh3:
- Replace Mountain-CAR problem with Atari problem (More complicated problem)
- Add implementation of DoubleDQN and DuelingDQN agent
- Add module to save and load model
- Update replay memory module


fluid/DeepQNetwork/DQN_agent.py
```diff
@@ -0,0 +1,188 @@
...
+
+class DQNModel(object):
+    def __init__(self, state_dim, action_dim, gamma, hist_len, use_cuda=False):
+        self.img_height = state_dim[0]
+        self.img_width = state_dim[1]
+        self.action_dim = action_dim
+        self.gamma = gamma
+        self.exploration = 1.1
+        self.update_target_steps = 10000 // 4
+        self.hist_len = hist_len
```
yu239: 'hist_len' should be better called 'context' or 'contexts_n' according to the convention?

fluid/DeepQNetwork/DQN_agent.py
```diff
@@ -0,0 +1,188 @@
...
+                   dtype='float32'), \
+               fluid.layers.data(
+                   name='isOver', shape=[], dtype='bool')
+
+    def _build_net(self):
+        state, action, reward, next_s, isOver = self._get_inputs()
+        self.pred_value = self.get_DQN_prediction(state)
+        self.predict_program = fluid.default_main_program().clone()
+
+        reward = fluid.layers.clip(reward, min=-1.0, max=1.0)
```
yu239-zz, line 46: It might be better to just clip the rewards outside Fluid programs, before being passed in.

fluid/DeepQNetwork/DQN_agent.py
```diff
@@ -0,0 +1,188 @@
...
+        # define program
+        self.train_program = fluid.default_main_program()
+
+        # fluid exe
+        place = fluid.CUDAPlace(0) if self.use_cuda else fluid.CPUPlace()
+        self.exe = fluid.Executor(place)
+        self.exe.run(fluid.default_startup_program())
+
+    def get_DQN_prediction(self, image, target=False):
+        image = image / 255.0
```
yu239-zz, line 77: Same as the above for rewards processing. Data pre-processing might be better done outside in pure Python code.

fluid/DeepQNetwork/DQN_agent.py
```diff
@@ -0,0 +1,188 @@
...
+
+        variable_field = 'target' if target else 'policy'
+
+        conv1 = fluid.layers.conv2d(
+            input=image,
+            num_filters=32,
+            filter_size=[5, 5],
+            stride=[1, 1],
+            padding=[2, 2],
```
yu239-zz, line 86: I am not quite sure about the details of the original network structure reported for DQN. Did they use padding at all?

fluid/DeepQNetwork/DQN_agent.py
```diff
@@ -0,0 +1,188 @@
...
+                sync_ops.append(sync_op)
+        sync_program = sync_program.prune(sync_ops)
+        return sync_program
+
+    def act(self, state, train_or_test):
+        sample = np.random.random()
+        if train_or_test == 'train' and sample < self.exploration:
+            act = np.random.randint(self.action_dim)
+        else:
+            if np.random.randint(100) == 0:
```
yu239-zz: np.random.random() < 0.01 is more straightforward

fluid/DeepQNetwork/atari.py
```diff
@@ -0,0 +1,160 @@
...
+        m = self.ale.getScreenRGB()
+        return m.reshape((self.height, self.width, 3))
+
+    def _current_state(self):
+        """
+        :returns: a gray-scale (h, w) uint8 image
+        """
+        ret = self._grab_raw_image()
+        # max-pooled over the last screen
+        ret = np.maximum(ret, self.last_raw_screen)
```
yu239-zz: We should better comment on why a max pool is needed over the current state and the previous state. Is it because of the missing frame issue?

fluid/DeepQNetwork/DQN_agent.py
yu239-zz, line 19: 'history' should be better called 'contexts' according to the RL naming convention.

fluid/DeepQNetwork/train.py
```diff
@@ -0,0 +1,188 @@
...
+def run_train_episode(agent, env, exp):
+    total_reward = 0
+    state = env.reset()
+    step = 0
+    while True:
+        step += 1
+        history = exp.recent_state()
+        history.append(state)
+        history = np.stack(history, axis=0)
+        action = agent.act(history, train_or_test='train')
```
yu239-zz: Here you use history to do prediction. But in eval_agent() and play.py you only use one frame? Why are they not consistent?

yu239-zz (changes requested):

fluid/DeepQNetwork/DQN_agent.py
TomorrowIsAnOtherDay: Yes, it is better to do data pre-processing outside, but doing it in GPU would be faster, especially fluid doesn't support async data-preprocessing currently.

fluid/DeepQNetwork/DQN_agent.py
TomorrowIsAnOtherDay: Yes, it is better to do data pre-processing outside, but doing it in GPU would be faster, especially fluid doesn't support async data-preprocessing currently.

fluid/DeepQNetwork/DQN_agent.py
TomorrowIsAnOtherDay: This architecture is slightly different from original network, but it takes fewer iterations to converge in  practice.

fluid/DeepQNetwork/DQN_agent.py
TomorrowIsAnOtherDay: Good suggestion.

fluid/DeepQNetwork/atari.py
TomorrowIsAnOtherDay: This code is copied from tensorpack, it may avoid missing frame issue here.
I will append  a comment on it.
https://github.com/zenghsh3/models/blob/3bf1d4b2b48e4239f4b420c8f3592d7d47e1bb33/fluid/DeepQNetwork/atari.py#L19

fluid/DeepQNetwork/DQN_agent.py
TomorrowIsAnOtherDay: Good suggestion.

fluid/DeepQNetwork/train.py
TomorrowIsAnOtherDay: Not only use one frame here.The env used in eval_agent() is wrapped by  **FrameStack** wrapper, which restores the recent frames in environment.And in training process, recent frames were taken care of by replay memory.
https://github.com/zenghsh3/models/blob/3bf1d4b2b48e4239f4b420c8f3592d7d47e1bb33/fluid/DeepQNetwork/atari_wrapper.py#L26

yu239-zz (approved):

PR #982, opened 2018-06-12
Distributed training for transformer, local training as default

xuezhong:


fluid/neural_machine_translation/transformer/train.py
```diff
@@ -266,152 +246,234 @@ def split_data(data, num_part=dev_count):
...
-            for batch_id, data in enumerate(test_data()):
+            loss_name=sum_cost.name,
+            main_program=train_progm,
+            build_strategy=build_strategy)
+
+        def test_context():
+            # Context to do validation.
+            test_program = train_progm.clone()
+            with fluid.program_guard(test_program):
+                test_program = fluid.io.get_inference_program([avg_cost])
```
panyx0718, line 332: Is the above few lines equal to ```test_program = train_program.clone(for_test=True)```?

PR #1023, opened 2018-07-02
Fix the max length in Transformer to count start and end tokens in

guoshengCS:
Fix the max length in Transformer to count start and end tokens in


kuke (approved):
LGTM

PR #1046, opened 2018-07-13
Add MKL-DNN Benchmarking to CRNN-CTC

Sand3r-:
Implements benchmarking facilities, as well as script folder with exemplary training and inference scripts.


Sand3r-:
If any further guidance is required, I'd be glad to help and answer any questions.

Sand3r-:
If any further guidance is required, I'd be glad to help and answer any questions.

fluid/ocr_recognition/ctc_reader.py
```diff
@@ -111,7 +116,7 @@ def reader():
 
         return reader
 
-    def infer_reader(self, img_root_dir=None, img_label_list=None):
+    def infer_reader(self, img_root_dir=None, img_label_list=None, cycle=False):
```
luotao1, line 123: please add the meaning of cycle:
```
:param cycle:
:type cycle:
```

fluid/ocr_recognition/scripts/infer_cpu.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+OMP_NUM_THREADS=56 KMP_AFFINITY=granularity=fine,compact,1,0 python ../infer.py \
```
luotao1: 1. `OMP_NUM_THREADS=56`, please do not set such large numbers, since a lot of machines do not have such large number of cores.
2. `KMP_AFFINITY=granularity=fine,compact,1,0`, this depends on machine itself. how about change it like
https://github.com/PaddlePaddle/Paddle/blob/develop/benchmark/fluid/run.sh
```
export MKL_NUM_THREADS=1
export OMP_NUM_THREADS=1
ht=`lscpu |grep "per core"|awk -F':' '{print $2}'|xargs`
if [ $ht -eq 1 ]; then # HT is OFF
    if [ -z "$KMP_AFFINITY" ]; then
        export KMP_AFFINITY="granularity=fine,compact,0,0"
    fi
    if [ -z "$OMP_DYNAMIC" ]; then
        export OMP_DYNAMIC="FALSE"
    fi
else # HT is ON
    if [ -z "$KMP_AFFINITY" ]; then
        export KMP_AFFINITY="granularity=fine,compact,1,0"
    fi
fi
```
3. why not use `--parallel=true`, ParallelExecutor is much faster than Executor. please adjust all related scripts: `infer_cpu.sh` and `train_cpu.sh`

fluid/ocr_recognition/README.md
```diff
@@ -113,6 +113,10 @@ data/test_images/00003.jpg
 ```
 env CUDA_VISIABLE_DEVICES=0 python ctc_train.py
 ```
+使用默认数据在CPU上训练:
+```
+env OMP_NUM_THREADS=<num_of_physical_cores> KMP_AFFINITY=granularity=fine,compact,1,0 python ctc_train.py --use_gpu False --parallel=False
```
luotao1: please adjust this command based on following comments.

fluid/ocr_recognition/scripts/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+To perform inference on the trained model using *CPU* with MKL-DNN, please run:
+```sh
+source infer_mkldnn.sh
+```
+
+To perform inference on the trained model using *GPU*, please run:
+
+```sh
+source infer_gpu.sh
```
luotao1: Could you combine above 6 scripts into two script (like `train.sh` and `infer.sh`) with different arguments? Thus, if we change the arguments like `batch_size` or `input_images_lists`, we should not change all these 6 scripts.
Maybe you can refer to https://github.com/PaddlePaddle/Paddle/blob/develop/benchmark/paddle/image/run_mkl_train.sh

luotao1 (commented):
Thanks for your work! And sorry for the late review.

fluid/ocr_recognition/scripts/infer_cpu.sh
Sand3r-: I have adjusted the scripts according to your suggestions when it comes down to points 1 and 2.
However, regarding point 3 - `--parallel=true` would admittedly increase the performance, but the problem with CPU version of ParallelExecutor is that it disallows batch sizes lower than number of cores, making it heavily machine dependent. Hence it is disabled by default.

fluid/ocr_recognition/scripts/README.md
Sand3r-: Thank you for your suggestion.
The sole purpose of scripts was to give an example of a command allowing for training and inference of a given topology (as described [here](https://github.com/PaddlePaddle/models/pull/1046/files#diff-5183f87f9d91d8b6234d09ccf43ad5e4R2)), while the same time keeping it as simple as possible. Making them more generic would kill the simplicity.

tensor-tang: Yes, we can understand your point, but we thought it would be very nice to use two scripts(one for training one for inference) to control all kernels(GPU,CPU, and MKLDNN), batch size and some other tuning parameters. That would make our validation work simple and can prevent some unnecessary tuning error by human factor.

Sand3r-: The scripts have been adjusted according to your request.

fluid/ocr_recognition/scripts/infer.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+export MKL_NUM_THREADS=1
+export OMP_NUM_THREADS=1
+export LD_LIBRARY_PATH=/usr/local/lib
```
luotao1: line 4 should be removed. In gpu docker images, `LD_LIBRARY_PATH` contain  cuda path. And why do you export `LD_LIBRARY_PATH`?

This comment is same for the train.sh.

fluid/ocr_recognition/scripts/infer.sh
```diff
@@ -0,0 +1,43 @@
...
+
+mode=$1 # gpu, cpu, mkldnn
+if [ "$mode" = "CPU" ] || [ "$mode" = "cpu" ]; then
```
luotao1: To make the script simply, you can choose only one of `CPU` or `cpu`.
```
if [ "$mode" = "CPU" ]; then
   XXX
elif [ "$mode" = "GPU" ]; then
   XXX
```

This comment is same for the train.sh.

fluid/ocr_recognition/scripts/train.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+export MKL_NUM_THREADS=1
+export OMP_NUM_THREADS=1
+export LD_LIBRARY_PATH=/usr/local/lib
+
+mode=$1 # gpu, cpu, mkldnn
+if [ "$mode" = "CPU" ] || [ "$mode" = "cpu" ]; then
+  use_gpu="False"
+  parallel="False"
```
luotao1: why do parallel="False" in CPU mode? It should be "True".

luotao1 (commented):
1. Could you adjust the python code format to pass the travis ci? https://travis-ci.org/PaddlePaddle/models/builds/409854951#L768
2. > but the problem with CPU version of ParallelExecutor is that it disallows batch sizes lower than the number of cores, making it heavily machine dependent

Thanks for your point. 

@chengduoZH Could you help see how to fix this problem?

@Sand3r- Could you use `Parallel=True` in default mode now. If batch sizes lower than the number of cores, you can report as a warning. The reason is：
- The CPU mode and MKLDNN mode should be the same parallel mode.
- Users and QA will use train.sh directly, it's hard for them to modify the CPU mode into parallel.

fluid/ocr_recognition/scripts/infer_cpu.sh
chengduoZH: @Sand3r- Currently, if you use `ParallelExecutor`, you should ensure that every CPU thread must have training data which means the `batch_size` must be greater than or equal to `CPU_NUM`.

Sand3r-:
Requested changes have been made.
Changes which cause travis build to fail weren't introduced by me, but I've fixed them anyway.

fluid/ocr_recognition/scripts/train.sh
```diff
@@ -0,0 +1,52 @@
...
+  if [ $core_num -gt $batch_size ]; then
+    echo "Batch size should be greater or equal to the number of 
+          available cores, when parallel mode is set to True."
+  fi
+  use_gpu="False"
+  save_model_dir="mkldnn_model"
+  export FLAGS_use_mkldnn=1
+else
+  echo "Invalid mode provided. Please use one of {GPU, CPU, MKLDNN}"
+  exit 1
```
luotao1: CPU: parallel=True
GPU: parallel=True
MKLDNN: parallel=False

luotao1 (commented):
Merge the latest changes from develop into this branch

Sand3r-:
I've rebased the branch against the develop, to avoid unnecessary commits.

fluid/ocr_recognition/scripts/train.sh
```diff
@@ -0,0 +1,55 @@
...
+  parallel="True"
+elif [ "$mode" = "GPU" ]; then
+  use_gpu="True"
+  save_model_dir="gpu_model"
+  parallel="True"
+elif [ "$mode" = "MKLDNN" ]; then
+  if [ $core_num -gt $batch_size ]; then
+    echo "Batch size should be greater or equal to the number of 
+          available cores, when parallel mode is set to True."
+  fi
```
luotao1, line 24: For MKLDNN, line 21-24 could be removed. You can fix it in next PR.

luotao1 (approved):
LGTM! Thanks very much!

PR #1054, opened 2018-07-18
should be zip instead of tar

daming-lu:


qingqing01 (approved):
@daming-lu Thanks very much.

PR #1061, opened 2018-07-20
Make the Transformer network configurations more flexible

guoshengCS:
Make the Transformer network configurations more flexible


kuke (approved):
LGTM

PR #1069, opened 2018-07-23
Fix the learning rate error caused by args.local in Transformer

guoshengCS:
Fix the learning rate error caused by args.local in Transformer


kuke (approved):
LGTM

PR #1073, opened 2018-07-25
add smoke model for superiving under CE

guochaorong:


fluid/mnist/model.py
```diff
@@ -0,0 +1,203 @@
...
+        train_avg_acc = accuracy.eval()
+        train_avg_loss = np.mean(every_pass_loss)
+        test_avg_acc = eval_test(exe, batch_acc, batch_size_tensor,
+                                 inference_program)
+
+        print("pass=%d, train_avg_acc=%f,train_avg_loss=%f, test_avg_acc=%f, elapse=%f" %
+              (pass_id, train_avg_acc, train_avg_loss, test_avg_acc, (pass_end - pass_start)))
+
+        with open("train_acc_factor.txt", 'a+') as f:
+            f.write("%s\n" % train_avg_acc)
```
panyx0718: do we need to format the output so it looks different from others?

panyx0718 (approved):

PR #1082, opened 2018-07-29
Adversarial jsma

dailydreamer:
Add [original JSMA paper](https://arxiv.org/abs/1511.07528) implementation trick, choosing two pixels when attack instead of one. Attack accuracy has little improve.
jsma0
jsma1


dailydreamer:
Already merge to Baidu/Advbox

PR #1202, opened 2018-08-29
【DeepQNetwork】Compatible with python3 is supported

TomorrowIsAnOtherDay:


fluid/DeepQNetwork/train.py
```diff
@@ -74,7 +74,7 @@ def get_player(rom, viz=False, train=False):
 
 def eval_agent(agent, env):
     episode_reward = []
-    for _ in tqdm(xrange(30), desc='eval agent'):
+    for _ in tqdm(range(30), desc='eval agent'):
```
velconia, line 77: we recommend to use ```six.moves.xrange(30)``` here

TomorrowIsAnOtherDay: Thanks for pointing out.
Performance between **range** and **xrange** could be ignored if we only range from 30.

velconia (approved):

PR #1204, opened 2018-08-29
Fix random.shuffle in face_detection

baiyfbupt:


velconia:
Did the ```np.random.seed``` be set here?

velconia (approved):

PR #1211, opened 2018-08-30
Add Fasterrcnn model

jerrywgz:


qingqing01 (approved):
Will refine code later.

PR #1226, opened 2018-09-05
fix inference script in ocr_recognition

wojtuss:
The `get_feeder_data()` method needs the `need_label` parameter set to `False` in inference for it to work.


luotao1 (approved):
LGTM

Chanbluky:
But it seems there is no parameter "need_label" in get_attention_feeder_for_infer function.

PR #1284, opened 2018-09-25
Some improvements in DAM config

kuke:
1. Make the number of channels configurable;
2. Avoid setting word embedding via feeding;
3. Change the default batch size in training.


sneaxiy (approved):
Excellent

PR #1318, opened 2018-09-28
Fix bug in attention calculation

ktlichkid:
Fix implementation of simple_attention to match the implementation of other frameworks.


guoshengCS (approved):

kuke (approved):
LGTM

PR #1326, opened 2018-09-29
Disable CE until new KPIs generated

ktlichkid:
The model has now changed. We need to temporarily disable this CE until we have the latest KPIs.


kuke (approved):
LGTM

PR #1354, opened 2018-10-10
Optimize Transformer performance

guoshengCS:
1. Reshape decoder output from 3D to 2D to use GEMM instead of BatchedGEMM in Transformer, which is faster and uses fewer memories.
2. Reuse input as the output of reshape_op in Transformer.


kuke (approved):
LGTM

PR #1367, opened 2018-10-18
Add Text Matching models on Quora

mapingshuo:
Add Text Matching models on Quora dataset: cDSSM，DecAtt, InferSent and SSE model.


fluid/text_matching_on_quora/configs/basic_config.py
```diff
@@ -0,0 +1,42 @@
+from __future__ import print_function
```
guru4elephant, line 1: you should add copy right license here

mapingshuo: Added copy right license : https://github.com/PaddlePaddle/models/pull/1367/commits/f421c0785a93b295833109733dc54bbd4c3420aa

fluid/text_matching_on_quora/models/infer_sent.py
```diff
@@ -0,0 +1,72 @@
...
+        bi_lstm_h = bi_lstm_layer(
+                        embed,
+                        rnn_hid_dim = self._config.rnn_hid_dim, 
+                        name='encoder')
+
+        bi_lstm_h = fluid.layers.dropout(bi_lstm_h, dropout_prob=self._config.droprate_lstm)
+        pool = fluid.layers.sequence_pool(input=bi_lstm_h, pool_type='max')
+        return pool
+
+    def MLP(self, vec):
```
guru4elephant, line 62: if MLP is frequently used in your model zoos, common functions should be implemented in a Base class for all Nets. it should be a Todo in the future to simplify Net.

guru4elephant (approved):
all models should be able to run on distributed environment in the next version. I think it is ok to stop here before v1.1

PR #1368, opened 2018-10-18
add RoIAlign for faster RCNN model

jerrywgz:


fluid/faster_rcnn/models/model_builder.py
```diff
@@ -196,12 +196,21 @@ def fast_rcnn_heads(self, roi_input):
...
-            pooled_height=14,
-            pooled_width=14,
-            spatial_scale=0.0625)
+        if cfg.roi_func == 'RoIPool':
+            pool = fluid.layers.roi_pool(
+                input=roi_input,
+                rois=pool_rois,
+                pooled_height=14,
+                pooled_width=14,
+                spatial_scale=0.0625)
```
xingyuanbu: pooled_width and spatial_scale should leave as a config in config.py
For example, pooled_height may take 7 or 14

jerrywgz: Configs are added

fluid/faster_rcnn/config.py
```diff
@@ -150,6 +150,12 @@
 # sampling ratio for roi align
 _C.sampling_ratio = 0
 
+# pooled width and pooled height 
+_C.roi_resolution = 14
+
+# spatial scale 
+_C.spatial_scale = 0.0625
```
xingyuanbu: 0.0625 is meanless, 1/16 would be better~

jerrywgz: Done

qingqing01 (approved):

PR #1417, opened 2018-11-01
 fix bug for L2DecayRegularizer

xuezhong:
L2DecayRegularizer only updates  SelectRows parameter which is hitted by input data. Fix the bug to get a  higher Rouge-L


kuke (approved):
LGTM

PR #1432, opened 2018-11-08
Fix the reshape in no-weight-sharing mode of Transformer

guoshengCS:
Fix the reshape in no-weight-sharing mode of Transformer


kuke (approved):
LGTM

PR #1489, opened 2018-11-30
fix tagspace gpu num_devices bug

frankwhzhang:


fluid/PaddleRec/tagspace/train.py
```diff
@@ -102,7 +102,7 @@ def train():
             if batch_id % args.print_batch == 0:
                 print("TRAIN --> pass: {} batch_num: {} avg_cost: {}, acc: {}"
                         .format(pass_idx, (batch_id+10) * batch_size, np.mean(loss_val),
-                                float(np.sum(correct_val)) / batch_size))
+                                float(np.sum(correct_val)) / (args.num_devices*batch_size)))
```
gavin1332, line 105: for one batch data will be uniformly divided into multi-GPU devices? if so, it looks good to me.

gavin1332 (approved):

PR #1571, opened 2018-12-27
Fix NLP-model for python3all: chinese_ner

junjun315:
1: fix xrange
2: fix `iteritems`
3: fix `map(xx lambda xx)`
4: fix numpy.type `isinstance` error ： Uncertainty of type after multiple operations


kuke (approved):
LGTM

PR #1643, opened 2019-01-11
fix previous merge bug

typhoonzero:
Since we use `softmax_with_cross_entropy` in train.py, need to remove fc output softmax activations

Fix changes from https://github.com/PaddlePaddle/models/pull/1533


Yancey0623 (approved):
lgtm

Issue #1647, opened 2019-01-14
faster_rcnn如果一旦数据指定错误，活找不到训练文件，程序会一直夯着


kolinwei:

jerrywgz:
请问有具体例子吗

jerrywgz:
在faster rcnn的目录下有data_utils.py中会检查如果数据读取失败会报错

jerrywgz:
在faster rcnn的目录下有data_utils.py，会检查如果数据读取失败会报错

kolinwei:
Creating: coco2017
loading annotations into memory...
Done (t=28.11s)
creating index...
index created!
_add_gt_annotations took 31.525s
Appending horizontally-flipped training examples...
Loaded dataset: coco2017
236574 roidb entries
Filtered 2042 roidb entries: 236574 -> 234532
train on coco2017 with 236574 roidbs
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/opt/python/cp27-cp27mu/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/opt/python/cp27-cp27mu/lib/python2.7/threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/opt/python/cp27-cp27mu/lib/python2.7/site-packages/paddle/fluid/layers/io.py", line 563, in __provider_thread__
    for tensors in func():
  File "/opt/python/cp27-cp27mu/lib/python2.7/site-packages/paddle/fluid/layers/io.py", line 610, in __tensor_provider__
    for slots in paddle_reader():
  File "/opt/python/cp27-cp27mu/lib/python2.7/site-packages/paddle/fluid/data_feeder.py", line 280, in __reader_creator__
    for item in reader():
  File "/workspace/work/d70dbdcf75ece5ee/modelce/tasks/faster_rcnn/reader.py", line 107, in reader
    roidb, mode)
  File "/workspace/work/d70dbdcf75ece5ee/modelce/tasks/faster_rcnn/reader.py", line 67, in roidb_reader
    im, im_scales = data_utils.get_image_blob(roidb, mode)
  File "/workspace/work/d70dbdcf75ece5ee/modelce/tasks/faster_rcnn/data_utils.py", line 47, in get_image_blob
    'Failed to read image \'{}\''.format(roidb['image'])
AssertionError: Failed to read image 'dataset/coco/train2018/000000428485.jpg'

PR #1666, opened 2019-01-17
fix hang when load error in faster_rcnn

jerrywgz:
https://github.com/PaddlePaddle/models/issues/1647

### DIFF
--- a/fluid/PaddleCV/faster_rcnn/data_utils.py
+++ b/fluid/PaddleCV/faster_rcnn/data_utils.py
@@ -28,6 +28,7 @@ from __future__ import unicode_literals
 import cv2
 import numpy as np
 from config import cfg
+import os
 
 
 def get_image_blob(roidb, mode):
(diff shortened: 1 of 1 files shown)


fluid/PaddleCV/faster_rcnn/data_utils.py
```diff
@@ -43,8 +44,9 @@ def get_image_blob(roidb, mode):
         target_size = cfg.TEST.scales[0]
         max_size = cfg.TEST.max_size
     im = cv2.imread(roidb['image'])
-    assert im is not None, \
-        'Failed to read image \'{}\''.format(roidb['image'])
+    if im is None:
+        print('Failed to read image \'{}\''.format(roidb['image']))
```
qingqing01, line 50: Please use raise, not print

https://stackoverflow.com/questions/2146618/raising-an-exception-vs-printing

jerrywgz: 已修改，是否要把这一系列问题提个需求

qingqing01 (approved):

PR #1746, opened 2019-02-12
add resume weights in train.py

heavengate:


heavengate:
> maybe --resume is better, since we resume not only weights but also other variables such as lr and lr count

Done

heavengate:
> fluid.io.load_vars

Done

fluid/PaddleCV/video/train.py
```diff
@@ -59,6 +60,12 @@ def parse_args():
         default=None,
         help='path to pretrain weights. None to use default weights path in  ~/.paddle/weights.'
     )
+    parser.add_argument(
+        '--resume',
+        type=str,
+        default=None,
+        help='path to resume weights. If not None, only resume weigths will be loaded.'
```
SunGaofeng: this comment may be changed, help='path to resume training based on previous checkpoints. If not None, training will be resumed, otherwise do nothing.'
since we not only load weights but also lr and lr decay count when resume.

SunGaofeng (approved):

PR #1796, opened 2019-02-27
Update reader.py

zzszmyf:
the _reader_creator  function needs default arg, if it not, the code can't work


zzszmyf (commented):
trainer_id of the _reader_creator functions need default value, if not, the code can't work.

PR #2036, opened 2019-04-12
fix - and _ mismatch, fix python3 bug in nonlocal codes

SunGaofeng:
1 fix - and _ mismatch in readme
2 fix bug of nonlocal codes in python3


heavengate (approved):
LGTM

PR #2043, opened 2019-04-12
Use new memory optimize API in image classfication

zhhsplendid:
The training script used a deprecated memory optimization API and some new memory optimization strategies are not used. If we use the new API, we are able to be same at GPU usage performance of PyTorch for SE-Resnext model. I changed the API to use the new API.


zhhsplendid:
Add QA evaluated test accuracy during training epochs. The training was done with 120 epochs under multi-GPU mode.

In the following figures, the curve "test_with_gc_" means the code adding memory optimization and "test_with_nogc_" means the code without this PR.

CF07CEB38B2826B3D53E69F34CE9FFE5

0DED564B5B4C91F6DF391A9CAAC2E9A5

zhhsplendid:
I will close this PR because the performance team will group our changes in v1.5 together.

PR #2044, opened 2019-04-12
add GoogleNet support add quant low level api usage doc.

wzzju:
* Fix the out node of GoogleNet.
* Add quantization low-level API usage doc.


wanghaoshuang (approved):
excellent

PR #2063, opened 2019-04-16
Enable GC and fix loading checkpoint in slim demo.

wanghaoshuang:
1. Enable GC strategy in run.sh
2. Fix loading checkpoint in filter pruning strategy by adding init_model option into config file.


wzzju (approved):
LGTM.

PR #2098, opened 2019-04-19
fix config for ce

jerrywgz:


jerrywgz:
fix https://github.com/PaddlePaddle/models/issues/2085

kolinwei (approved):

PR #2103, opened 2019-04-21
Update .run_ce.sh

kolinwei:


jerrywgz (approved):
Sry for type-error
LGTM

PR #2136, opened 2019-04-23
Modify single queue to a queue per process for data processing

sandyhouse:
Use a queue for each image processing process to improve performance.


guru4elephant (approved):
LGTM

PR #2238, opened 2019-05-14
[Bug fix] fix overflow error in rcnn

jerrywgz:
For faster rcnn in Windows,  overflow error is catched
 https://github.com/PaddlePaddle/models/issues/2216


heavengate (approved):
LGTM

PR #2252, opened 2019-05-15
Add a note in image classification to remind users to replace PE to exec 

shippingwang:
Because WINDOWS does not support NCCL. It will raise an error when users try to run the multi-GPUs program on WINDOWS machine. So add a note to remind users to replace parallel executor to executor.


SunGaofeng (approved):
LGTM

PR #2254, opened 2019-05-16
Refine PE doc in rcnn for Windows users

jerrywgz:
Because WINDOWS does not support NCCL. It will raise an error when users use parallel executor on WINDOWS machine. So add a note to remind users to replace parallel executor to executor.


heavengate (approved):
LGTM

PR #2267, opened 2019-05-16
test=develop

xyzhou-puck:
Add two examples of using paddle.fluid.dygraph to implement reinforcement learning models.

Including:

reinforce.py
actor_critic.py

Both two models pass the Cartpole-v0 (and Cartpole-v1) games.


phlrain (approved):
LGTM

PR #2315, opened 2019-05-27
add pretrain related statement in readme

SunGaofeng:
add description in readme to show that some models should be trained with pertained models.


heavengate (approved):
LGTM

PR #2320, opened 2019-05-28
Update mnist dygraph

chengduoZH:
In multi-process mode, the method of getting Place should be :
```
place = fluid.CUDAPlace(fluid.dygraph.parallel.Env().dev_id) \
        if args.use_data_parallel else fluid.CUDAPlace(0)
```


JiabinYang (approved):
LGTM

PR #2344, opened 2019-06-02
fix error while loading cv2.VideoCapture

cjt222:
1、fix error which cv2 is used but not import while loading mp4 file
2、while loading mp4 file, some frames may fail to be read and it may result in out of range in sampledFrames
image


SunGaofeng (approved):

SunGaofeng:
Good, thanks to your pr
1,  import cv2 
2, replace videolen by len(sampledFrames)

PR #2361, opened 2019-06-05
init rcnn mixed precision training

Jie-Fang:
This feature implements mixed precision training on rcnn. It utilizes auto loss scaling, black white lists to make training stable. 
Testing this feature requires this [pull request](https://github.com/PaddlePaddle/Paddle/pull/17847) in Paddle and turns on `--use_fp16=True` and `use_dynamic_loss_scaling=True`. You can also set `--loss_scaling`, `--incr_every_n_steps`, `--decr_every_n_nan_or_inf`, `--incr_ratio`, `--decr_ratio` to control the detail.

### DIFF
--- a/PaddleCV/rcnn/models/load_model.py
+++ b/PaddleCV/rcnn/models/load_model.py
@@ -1,0 +1,95 @@
+import os
+from paddle.fluid import layers
+import paddle.fluid.core as core
+from paddle.fluid.executor import Executor
+from paddle.fluid.evaluator import Evaluator
+from paddle.fluid.framework import Program, Parameter, default_main_program, default_startup_program, Variable, program_guard
+
+def _clone_var_in_block_(block, var, name, dtype):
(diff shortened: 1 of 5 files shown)


qingqing01:
Thanks! We try this to verify accuracy.

PaddleCV/rcnn/models/load_model.py
```diff
@@ -0,0 +1,95 @@
+import os
```
qingqing01, line 1: It seems that it is not very friendly to let users to write these things.

Jie-Fang: Yes, I think so. The pretrained models are fp32, but we need fp16 weight copy to initialize fp16 weights in mixed precision. Thus I modify `load_vars` function slightly. The best way is to add this feature in Paddle. My implementation is a temporary plan for testing rcnn mixed precision training. The better implementation should be discussed.

PR #2363, opened 2019-06-05
fix bug of gan in python3

ceci3:
fix bug of gan in above issues:
https://github.com/PaddlePaddle/models/issues/2356
https://github.com/PaddlePaddle/models/issues/2355


SunGaofeng (approved):
approved

PR #2384, opened 2019-06-11
Update the emotion demo and use the PaddleHub to get the ERNIE model

nepeplwu:


chenbjin (approved):
emotion_detection 使用PaddleHub接入ERNIE预训练模型，验证没问题

PR #2386, opened 2019-06-11
modify on nextvald, tsm and nonlocal to support v1.5

SunGaofeng:
modifies include:
1, nextvald, move floating data processing into fluid to process on gpu
2, tsm, add gc FLAGS in train_tsm.sh to enable large batchSize
3, nonlocal, add c2d resnet101 and i3d rennet50 support.

### DIFF
--- a/PaddleCV/video/train.py
+++ b/PaddleCV/video/train.py
@@ -140,9 +140,6 @@ def train(args):
             optimizer.minimize(train_loss)
             train_pyreader = train_model.pyreader()
 
-    if not args.no_memory_optimize:
-        fluid.memory_optimize(train_prog)
-
     valid_prog = fluid.Program()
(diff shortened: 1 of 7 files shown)


PaddleCV/video/train.py
```diff
@@ -140,9 +140,6 @@ def train(args):
             optimizer.minimize(train_loss)
             train_pyreader = train_model.pyreader()
 
-    if not args.no_memory_optimize:
```
shippingwang, line 143: maybe should remove args: no_memory_optimize

shippingwang (approved):
LGTM

PaddleCV/video/train.py
SunGaofeng: yes, if this can be confirmed no need in future codes.

PR #2435, opened 2019-06-18
 fix load bug for cycle_gan inference 

DDDivano:
load_persistables function returns a tuple which include parameters and optimizer


tink2123:
LGTM! thanks~

ceci3 (approved):

PR #2474, opened 2019-06-21
Fix cycle gan cudnn algorithm bug

sneaxiy:
`FLAGS_exhaustive_search` is conflict with `FLAGS_cudnn_deterministic`, which makes CE fail. This PR fixes it.


wanghaoshuang (approved):
LGTM

PR #2490, opened 2019-06-21
fix some bug of gan

ceci3:
add else for data_reader.py
delete some print for starGAN

### DIFF
--- a/PaddleCV/gan/data_reader.py
+++ b/PaddleCV/gan/data_reader.py
@@ -262,7 +262,7 @@ class celeba_reader_creator(reader_creator):
         attr_names = args.selected_attrs.split(',')
         for line in lines:
             arr = line.strip().split()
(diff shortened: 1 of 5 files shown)


PaddleCV/gan/data_reader.py
```diff
@@ -262,7 +262,7 @@ def __init__(self,
         attr_names = args.selected_attrs.split(',')
         for line in lines:
             arr = line.strip().split()
-            name = './images/' + arr[0]
+            name = './img_align_celeba/' + arr[0]
```
qingqing01: use  `os.path.join`

qingqing01: 需要避免这种hard code

ceci3: done thanks

qingqing01 (approved):

PR #2531, opened 2019-06-25
fix no random bug

AIFollowers:
the use_random of generate_proposal_label op have bug, can't be configed from cfg.


willthefrog:
`shuffle_before_sample=True, `，这个默认值会变成tuple吗？还是逗号就忽略了？

willthefrog:
also this one `is_cascade_rcnn=True if curr_stage > 0 else False, `

qingqing01 (approved):

PR #2538, opened 2019-06-25
add ce for neural_machine_translation/transformer

zhengya01:
add ce for neural_machine_translation/transformer


guoshengCS (approved):
LGTM

PR #2571, opened 2019-06-27
Fix ptb lm forget bias

phlrain:
Fix ptb lm forget bias

set forget bias =0.0


JiabinYang (approved):
LGTM, Better to avoid using key words of Python

PR #2585, opened 2019-06-28
Clean up config files

willthefrog:
- reorder: batch_size -> data_set -> sample_transforms -> batch_transforms -> worker -> other
- unify bool values to lower case true/false
- remove `null` values
- remove `shuffle` settings, covered by default values
- remove `use_padded_im_info` settings, covered by default value
- update ResNet50 pretrained weight url

PR #2586, opened 2019-06-28
unify True/False in yml

heavengate:
**use True/False instead of true/false in configs**
some use true/false and others use True/False
since configs can be generate by python code, use python style True/False


heavengate:
unified in https://github.com/PaddlePaddle/models/pull/2585

PR #2618, opened 2019-06-29
refine model zoo doc

jerrywgz:
Refine model zoo doc and add more urls in models


jerrywgz:
MAP in SSD and YOLOV3 need to be fixed.

qingqing01 (approved):

PR #2656, opened 2019-07-01
modified acc calculation for se_resnext

tink2123:
- rm redundant code
- modified acc calculation for se_resnext


phlrain (approved):
LGTM

PR #2681, opened 2019-07-02
fix subprocess not stop when press ctr+c to stop main process

SunGaofeng:
register signal.SIGINT to fix subprocess not stop when press ctr+c to stop main process


heavengate (approved):
LGTM

SunGaofeng:
fix subprocess not terminate bug when press ctr+c

PR #2724, opened 2019-07-04
add check before using cuda

zhangyimi:


PaddleNLP/similarity_net/run_classifier.py
```diff
@@ -454,6 +457,15 @@ def main(conf_dict, args):
 
 if __name__ == "__main__":
     utils.print_arguments(args)
+    try:
+        if fluid.core.is_compiled_with_cuda() != True and args.use_cuda == True:
```
xyzhou-puck: please avoid using fluid.core because fluid.core is not designed for paddle users.

PR #2732, opened 2019-07-05
add cuda check to show errors when run on paddle compiled with cpu only

SunGaofeng:
Sometimes user would install cpu only paddle. So when use_gpu=True, the program should check whether paddle was compiled with gpu or cpu only. For the latter case, program will exit will error information to show that.

### DIFF
--- a/PaddleCV/PaddleVideo/utils.py
+++ b/PaddleCV/PaddleVideo/utils.py
@@ -40,8 +40,8 @@ class AttrDict(dict):
             self[key] = value
 
 def check_cuda(use_cuda, err = \
-    "\nYou can not set use_cuda = True in the model because you are using paddlepaddle-cpu.\n \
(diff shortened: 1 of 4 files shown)


PaddleCV/PaddleVideo/utils.py
```diff
@@ -36,3 +38,14 @@ def __setattr__(self, key, value):
             self.__dict__[key] = value
         else:
             self[key] = value
+
+def check_cuda(use_cuda, err = \
+    "\nYou can not set use_cuda = True in the model because you are using paddlepaddle-cpu.\n \
```
heavengate: use_cuda=True -> use_gpu=True ?

SunGaofeng: Yes, modified,  use_cuda to use_gpu

heavengate (approved):
LGTM

PR #2739, opened 2019-07-05
cherry pick to 1.5 for check cuda and python3 modifications

SunGaofeng:
1, add check cuda when use gpu
2, fix ctcn_reader to fit both python3 and 2 in pickle.load
3, fix program not terminate when ctr+c pressed


shippingwang (approved):

heavengate (approved):
LGTM

PR #2779, opened 2019-07-10
fix faster rcnn use im_shape

heavengate:
**fix save_inference_model cannot find im_shape in faster rcnn**
image


jerrywgz (approved):
LGTM

PR #2781, opened 2019-07-10
refine bbox_normalize in infer.py

heavengate:
**Refine is_bbox_normalized check in PaddleDetection infer.py**
1. add is_bbox_normalized() in SSD
2. tools/infer.py get is_bbox_normalized from model

### DIFF
--- a/PaddleCV/PaddleDetection/tools/infer.py
+++ b/PaddleCV/PaddleDetection/tools/infer.py
@@ -164,6 +164,13 @@ def main():
     clsid2catid, catid2name = get_category_info(anno_file, with_background,
                                                 use_default_label)
 
+    # only SSD bbox output is normalized values in range [0, 1]
+    is_bbox_normalized = False
+    if 'bbox' in keys:
+        bbox_var_name = values[keys.index('bbox')]
+        if bbox_var_name.find('detection_output') >= 0:
+            is_bbox_normalized = True
+
     imid2path = reader.imid2path
     for iter_id, data in enumerate(reader()):
         outs = exe.run(infer_prog,
@@ -178,7 +185,6 @@ def main():
 
         bbox_results = None
         mask_results = None
-        is_bbox_normalized = True if cfg.metric == 'VOC' else False
         if 'bbox' in res:
             bbox_results = bbox2out([res], clsid2catid, is_bbox_normalized)
         if 'mask' in res:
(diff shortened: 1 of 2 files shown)


heavengate:
since is_bbox_normalized is determined by output layer and is not configurable, make it as a attribute of models, and only SSD is_bbox_normalized = True.
please help to review~ @qingqing01 @willthefrog

PaddleCV/PaddleDetection/ppdet/modeling/architectures/base.py
```diff
@@ -0,0 +1,44 @@
...
+from __future__ import division
+from __future__ import print_function
+
+
+class ArchitectureBase(object):
+    """
+    Base object for architectures.
+    """
+
+    __category__ = 'architecture'
```
willthefrog: not needed as it is not registered

willthefrog:
IMHO, a base class seems a bit overkill for just this purpose, adding a attribute (or class attribute) and checking with `hasattr` should suffice.

Currently the model  interfaces are defined by convention, which is more pythonic. Interface by inheritance (a la java) does not add much as most of the methods are one-liners any way.

In my initial design there is only a `build(feed_var, mode)` method, all these `train`/`test` /`eval` are just shortcuts for backward compatibility during the refactor. Come to think about it now, these shortcuts should be simply removed.

heavengate:
> IMHO, a base class seems a bit overkill for just this purpose, adding a attribute (or class attribute) and checking with `hasattr` should suffice.
> 
> Currently the model interfaces are defined by convention, which is more pythonic. Interface by inheritance (a la java) does not add much as most of the methods are one-liners any way.
> 
> In my initial design there is only a `build(feed_var, mode)` method, all these `train`/`test` /`eval` are just shortcuts for backward compatibility during the refactor. Come to think about it now, these shortcuts should be simply removed.

Thanks, done, please help to review~

PaddleCV/PaddleDetection/tools/infer.py
```diff
@@ -164,6 +164,11 @@ def main():
     clsid2catid, catid2name = get_category_info(anno_file, with_background,
                                                 use_default_label)
 
+    # whether output bbox is normalized in model output layer
+    is_bbox_normalized = False
+    if hasattr(model, 'is_bbox_normalized'):
```
willthefrog: check callable? or simply use a member variable instead of method?

heavengate: Done, thanks

willthefrog (approved):
LGTM

PR #2790, opened 2019-07-11
Prevent module instance from modifying global config

willthefrog:
some of the global config values are reference type, e.g., objects, lists or
dicts, if the created module instances modify those values (see FPN and
spatial_scale), global config will reflect these changes, and instances of the
same class created later will inherit the changed values


jerrywgz (approved):
LGTM, thanks

PR #2791, opened 2019-07-11
Fixes command to check paddle paddle installation

tachyon77:


qingqing01:
@tachyon77 Thanks very much!   also fixed https://github.com/PaddlePaddle/models/pull/2741

PR #2794, opened 2019-07-12
Fix ctr startup run bug

sneaxiy:
Startup program should be run before pe constructs. It is because parameters would be broadcast from device 0 to other devices when pe constructs.


guru4elephant (approved):
LGTM

PR #2823, opened 2019-07-16
Fix the defalut value of CPU_NUM to be consistent with Paddle framework.

guoshengCS:
Fix the default value of CPU_NUM to be consistent with Paddle framework.
Fix the leaving out of subprocess importing.


kuke (approved):
LGTM

PR #2847, opened 2019-07-18
fix_train_eval

jerrywgz:
1. Fix duplicated name in Mask-RCNN when train and eval in same time.
2. Add create(model) at stage of eval.
3. Fix output path in faster_rcnn_r50_fpn_1x.yml.


heavengate (approved):
LGTM

Issue #2559, opened 2019-06-26
window上lexical_analysis infer报错


JiaXiao243:
在PaddleNLP/lexical_analysis目录下，执行python run_sequence_labeling.py --do_train False --do_test False --do_infer True --batch_size 80 --word_emb_dim 768 --grnn_hidden_dim 768 --bigru_num 2   --use_cuda True  --init_checkpoint ./model_baseline/ --infer_data ./data/test.tsv --word_dict_path ./conf/word.dic --label_dict_path ./conf/tag.dic  --word_rep_dict_path ./conf/q2b.dic命令，对该模型进行预测，报错信息如下:
![image](https://user-images.githubusercontent.com/37854899/60176210-37a2e380-9848-11e9-88a4-cee89e6ce14b.png)

PR #2849, opened 2019-07-18
fix print issue on win for py2

wopeizl:
test=develop
fix the print issue for python2 on windows
unicode character may lead to bad behavior if do print opertion on windows


wopeizl:
https://github.com/PaddlePaddle/models/issues/2559

kuke (approved):
LGTM

PR #2873, opened 2019-07-22
fix gt_id overflow and add README for windows

heavengate:
1.  remove unused gt_id, which may overflow in int32
2. add README for setting --use_multiprocess=False on Windows


jerrywgz (approved):
LGTM

PR #2886, opened 2019-07-22
fix yolov3 subprocess not exit

heavengate:
**fix yolov3 reader process not exit after main process killed**


SunGaofeng (approved):
approved

PR #2896, opened 2019-07-23
Code for the paper "robust neural machine translation with joint textual and phonetic embedding""

AndyELiu:
This is the migrated code for the paper "robust neural machine translation with joint textual and phonetic embedding"


kuke (approved):
LGTM

PR #2911, opened 2019-07-24
add multiscale-training

jerrywgz:
- Add multiscale-training. To align Detectron, multiscale-training is used for each image
- Fix ResizeImage using PIL interpolation method
- Enable evaluate model_final at the same time of training

### DIFF
--- a/PaddleCV/PaddleDetection/tools/train.py
+++ b/PaddleCV/PaddleDetection/tools/train.py
@@ -171,8 +171,9 @@ def main():
             it, np.mean(outs[-1]), logs, end_time - start_time)
         logger.info(strs)
 
-        if it > 0 and it % cfg.snapshot_iter == 0:
-            checkpoint.save(exe, train_prog, os.path.join(save_dir, str(it)))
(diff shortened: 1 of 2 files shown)


PaddleCV/PaddleDetection/tools/train.py
```diff
@@ -171,8 +171,9 @@ def main():
             it, np.mean(outs[-1]), logs, end_time - start_time)
         logger.info(strs)
 
-        if it > 0 and it % cfg.snapshot_iter == 0:
-            checkpoint.save(exe, train_prog, os.path.join(save_dir, str(it)))
+        if it > 0 and it % cfg.snapshot_iter == 0 or it == cfg.max_iters - 1:
```
qingqing01, line 174: why change here

jerrywgz: Now model_final is not evaluated when training and eval at same time. This change enables this case.

qingqing01 (approved):

PR #2917, opened 2019-07-24
fix infer doc and config default

heavengate:
1. fix infer doc `--output_dir`
2. fix TestFeed config annotation default path
3. refine auto download trigger only when `dataset_dir` end with `coco` or `voc`
4. refine dataset exists condition

### DIFF
--- a/PaddleCV/PaddleDetection/ppdet/utils/download.py
+++ b/PaddleCV/PaddleDetection/ppdet/utils/download.py
@@ -67,14 +67,13 @@ def get_weights_path(url):
     return get_path(url, WEIGHTS_HOME)
 
 
-def get_dataset_path(path):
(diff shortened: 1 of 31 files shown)


PaddleCV/PaddleDetection/ppdet/utils/download.py
```diff
@@ -82,7 +82,7 @@ def get_dataset_path(path):
                     osp.realpath(path), DATASET_HOME))
 
     for name, dataset in DATASETS.items():
-        if path.lower().find(name) >= 0:
+        if os.path.split(path.strip().lower())[-1] == name:
```
jerrywgz, line 84: 这里为什么要这样判断呢？

jerrywgz (approved):
LGTM

PR #2995, opened 2019-08-01
fix pred in simnet

Lizhengo:
修改simnet在训练过程中valid=False，但test=True时的bug, issue:https://github.com/PaddlePaddle/models/issues/2968


xixiaoyao (approved):

kuke (approved):
LGTM

PR #2996, opened 2019-08-01
fix windows multiprocess

heavengate:
multiprocess reader not compatible on Windows, use multi-thread instread of multi-process on Windows


jerrywgz (approved):
LGTM

PR #3000, opened 2019-08-01
Fix dygraph transformer

JiabinYang:
This PR will fix dropout seed to 0 instead of None, Since #18813 will cause None type of dropout seed crashed


DDDivano (approved):
LGTM

PR #3035, opened 2019-08-06
Use gc, PyReader & compiledprogram for bert

kuke:
As title


iclementine (commented):
It is fine from my point of view.

iclementine (approved):

PR #3053, opened 2019-08-08
fix yolov3 voc eval in train

heavengate:
**fix yolov3 eval in train on VOC dataset**
1. add voc_eval support zero padded Tensor
2. change yolov3 eval and eval in train read gt as zero padded tensor

**Tested on:** python2/3, yolo_coco, yolo_voc, ssd_voc, faster/mask/cascade rcnn, retinamet


jerrywgz (approved):
LGTM

PR #3071, opened 2019-08-09
refine weight download

heavengate:
**refine weight download**
download to tmp filename firstly and move to final filename after download finished.


jerrywgz (approved):
LGTM

PR #3102, opened 2019-08-13
Update test_architectures.py

wonyoungsen:
Correct the path of the ssd ".yml" file


qingqing01:
@wonyoungsen  Thanks very much! We don't noticed this PR before. And This  issue has been fixed.

PR #3131, opened 2019-08-19
fix enable_inplace

heavengate:
1. set default enable_inplace=False in train.py
2. refine image_path support windows stype path


jerrywgz (approved):
LGTM

PR #3168, opened 2019-08-22
Enhance the training processing of rnn_rearch

Xreki:
- Add an option to use parallel_executor
- Print more information
- Support the profile


phlrain (approved):
LGTM

PR #3229, opened 2019-08-30
fix bug in sharedmemory

walloollaw:
updates:
1, fix bug in shared memory when allocating not continouse pages
2, expose `memsize` parameter for XXXFeed


PaddleCV/PaddleDetection/ppdet/data/data_feed.py
```diff
@@ -299,6 +296,7 @@ def __init__(self,
                  num_workers=2,
                  bufsize=10,
                  use_process=False,
+                 memsize=None,
```
qingqing01, line 303: add comments for memsize in Args

walloollaw: done

qingqing01 (approved):

PR #3232, opened 2019-08-30
Add distributed training support to ppdet

willthefrog:
some code could be pretty hacky, due to time and other limitations  🤷‍♂️
current reader does not support partition (status may change very soon), so each worker will loop through the whole dataset,  make sure there are enough iterations.

### DIFF
--- a/PaddleCV/PaddleDetection/tools/train.py
+++ b/PaddleCV/PaddleDetection/tools/train.py
@@ -54,6 +54,14 @@ logger = logging.getLogger(__name__)
 
 
 def main():
+    env = os.environ
+    FLAGS.dist = 'PADDLE_TRAINER_ID' in env and 'PADDLE_TRAINERS_NUM' in env
+    if FLAGS.dist:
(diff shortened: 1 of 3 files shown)


PaddleCV/PaddleDetection/tools/train.py
```diff
@@ -136,15 +148,42 @@ def main():
...
+            trainer_id=trainer_id)
+
+    else:
+        exe.run(startup_prog)
+        pe = fluid.ParallelExecutor(
+            main_program=train_prog,
+            use_cuda=True,
+            loss_name=loss.name,
+            build_strategy=build_strategy)
+
```
qingqing01, line 183: 需要参考image_classification里的代码简化这块的逻辑

willthefrog: The model guide version does not  work, I suggest we merge this for now since it is working code. The style issue could be addressed later after all the major changes have been merged.

PR #3237, opened 2019-08-31
add group normalization head to cascade mask rcnn

Noplz:
add group normalization head to cascade mask rcnn
ref:https://github.com/zhaoweicai/Detectron-Cascade-RCNN/blob/master/detectron/modeling/cascade_rcnn_heads.py


jerrywgz (approved):
LGTM

PR #3260, opened 2019-09-04
add config map_type to set voc_eval map_type

heavengate:
**Add config map_type to set voc_eval map_type**
```yml
metric: VOC
map_type: 11point
```


jerrywgz (approved):
LGTM

PR #3273, opened 2019-09-05
fix add map_type

heavengate:
**fix config map_type**
In COCO metric map_type may not be set, fix cfg.map_type error


jerrywgz (approved):
LGTM

PR #3292, opened 2019-09-08
Release auto mixed-precision training for bert

kuke:
Update BERT for the fp16 support.

### DIFF
--- a/PaddleNLP/language_representations_kit/BERT/optimization.py
+++ b/PaddleNLP/language_representations_kit/BERT/optimization.py
@@ -19,7 +19,7 @@ from __future__ import print_function
 
 import numpy as np
 import paddle.fluid as fluid
-from utils.fp16 import create_master_params_grads, master_param_to_train_param
+from utils.fp16 import create_master_params_grads, master_param_to_train_param, apply_dynamic_loss_scaling
 
 
 def linear_warmup_decay(learning_rate, warmup_steps, num_train_steps):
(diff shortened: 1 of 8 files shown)


PaddleNLP/language_representations_kit/BERT/optimization.py
```diff
@@ -19,7 +19,7 @@
 
 import numpy as np
 import paddle.fluid as fluid
-from utils.fp16 import create_master_params_grads, master_param_to_train_param
+from utils.fp16 import create_master_params_grads, master_param_to_train_param, apply_dynamic_loss_scaling
```
gongweibao, line 22: Use decorate.

PaddleNLP/language_representations_kit/BERT/utils/fp16.py
```diff
@@ -17,26 +17,20 @@
 import paddle.fluid as fluid
 
 
-def cast_fp16_to_fp32(i, o, prog):
+def append_cast_op(i, o, prog):
```
gongweibao, line 20: These files should not exists.

gongweibao (changes requested):

iclementine (approved):
Looks Good to Me!

PaddleNLP/language_representations_kit/BERT/optimization.py
kuke: The decorate interface should be refined, before that we use the customized code here.

PR #3296, opened 2019-09-09
modify readme for not support windows cpu environment

JesseyXujin:
add readme note for not support windows cpu environment.


xyzhou-puck (approved):
LGTM.

PR #3305, opened 2019-09-10
Make exec_strategy.num_iteration_per_drop_scope = 1 in detection models

sneaxiy:
The default value of `exec_strategy.num_iteration_per_drop_scope` is 100 since [#Paddle19075](https://github.com/PaddlePaddle/Paddle/pull/19075). But in detection models such as `mask_rcnn_r101_fpn_1x.yml`, the garbages in scopes may be larger than 4G. This PR sets `exec_strategy.num_iteration_per_drop_scope` to be 1 to avoid memory leak.

42540804212a633682f2bf3eacace728


jerrywgz (approved):
LGTM

PR #3323, opened 2019-09-14
fix multi-process reader not exist in resume

heavengate:
**fix multi-process reader not exist in resume**
In resume mode, reader ```max_iter``` should be ```cfg.max_iter - start_iter```


jerrywgz (approved):
LGTM

PR #3364, opened 2019-09-19
yolov3 reader add except log

heavengate:
**yolov3 reader add except log**
log out reader exception, easier to debug


jerrywgz (approved):
Done

PR #3369, opened 2019-09-19
Fix dygraph model save

chengduoZH:
Currently, data parallel training using dygraph is a multi-process method, so if parameters need to be saved during training, the saving action should only be called in one process.


JiabinYang (commented):
maybe Add more desc in PR info?

JiabinYang (approved):
LGTM

PR #3374, opened 2019-09-20
Fix transformer inference when using ParallelExecutor

guoshengCS:
Fix transformer inference when using ParallelExecutor by setting Execution Strategy to clear tensor array after each iteration.


xyzhou-puck (approved):
LGTM.

PR #3446, opened 2019-09-29
Fix slim resnet prefix

baiyfbupt:
1. We do not need to add split character '_' when prefix_name is None.
2. Fix data path in the Slim classification distillation demo.


ceci3 (approved):
LGTM

PR #3452, opened 2019-09-29
Update ppdet/utils/checkpoint.py to skip the var shape check when the var is not initialized.

qingqing01:


willthefrog:
LGTM

willthefrog (approved):
LGTM

PR #3480, opened 2019-10-08
Move paddle.batch and paddle.reader.xxx to fluid.io

sneaxiy:
As discussed with @lanxianghit , all models should run under at least PaddlePaddle 1.6+.

`paddle.batch` and `paddle.reader.xxx` have moved to `fluid.io`. This PR removes the legacy usages to `fluid.io.xxx`.


wanghaoshuang (approved):
LGTM.

PR #3518, opened 2019-10-12
Add prefix_name in PaddleDetection

baiyfbupt:
Add a shared config `weight_prefix_name`, it can add a unified prefix in model parameters name.


PaddleCV/PaddleDetection/ppdet/core/workspace.py
```diff
@@ -125,7 +125,7 @@ def register(cls):
     return cls
 
 
-def create(cls_or_name, **kwargs):
+def create(cls_or_name, prefix_name='', **kwargs):
```
qingqing01: better to put weight_prefix_name in the YAML config, then pass the args by __init__, not this `create` interface

baiyfbupt: Fixed.

qingqing01 (approved):

PR #3524, opened 2019-10-12
update readme for paddle version

jerrywgz:
Add paddlepaddle version requirement for Paddle models and PaddleDetection


lanxianghit (approved):
LGTM

PR #3530, opened 2019-10-12
add version checking

hutuxian:
Some APIs are upgraded in PaddlePaddle version 1.6, so we add version checking for DIN and SR-GNN to remind user to upgrade PaddlePaddle.


jerrywgz (approved):
Need some description for this pr

PR #3547, opened 2019-10-14
Use new save/load api and change fluid.layers.data to fluid.data

lfchener:
Use new save/load api and change fluid.layers.data to fluid.data.


kuke (approved):
LGTM

PR #3550, opened 2019-10-14
Revert "unify reader to dataloader in language model"

lfchener:
Reverts PaddlePaddle/models#3488，Revert "unify reader to dataloader in language model"


iclementine (approved):
LGTM

PR #3561, opened 2019-10-14
Fix doc of pruing yolov3 demo in PaddleDetection.

wanghaoshuang:
Fix doc of pruing yolov3 demo in PaddleDetection.


baiyfbupt (approved):
LGTM

PR #3562, opened 2019-10-14
fix api of fluid.layers.data and fluid.io.PyReader

SunGaofeng:
fix api of fluid.layers.data and fluid.io.PyReader to fit release1.6
fluid.layers.data -> fluid.data
fluid.io.PyReader -> fluid.io.DataLoader.from_generator


ceci3 (approved):
LGTM

PR #3565, opened 2019-10-14
add ETS model to release 1.6

huangjun12:
ets is a video caption model, newly added to PaddlePaddle/models


SunGaofeng (approved):
approved

PR #3598, opened 2019-10-16
Paddingernie16

kinghuin:
1. Support Padding lac based on ernie
2. Support new unpad op in paddle1.6
3. fix input_mask dtype


kuke (approved):
LGTM

PR #3599, opened 2019-10-16
Upgrade bert to 1.6 api

kuke:
1. Use `fluid.data()` and `fluid.embedding()`
2. Remove `fluid.contrib.memory_usage()` because there are two unknown dimensions now and it cannot work any more.


xyzhou-puck (approved):
LGTM.

PR #3617, opened 2019-10-16
remove actual_shape in resize_nearest and use_progam_cache in yolov3/infer.py

heavengate:
1. `scale` parameter for `resize_nearest` change to scale input dynamically since Paddle 1.5, change to use scale to upsample feature maps in yolov3 and fpn
2. set `use_program_cache=True` in yolov3/infer.py, see https://github.com/PaddlePaddle/models/issues/3571


jerrywgz (approved):
LGTM

PR #3634, opened 2019-10-17
update api and add check_version

shippingwang:
- add PaddlePaddle version checking function
- update new data loader instead of py_reader
- use fluid.data interface instead of fluid.layers.data

### DIFF
--- a/PaddleCV/image_classification/train.py
+++ b/PaddleCV/image_classification/train.py
@@ -180,8 +180,8 @@ def train(args):
     train_reader = imagenet_reader.train(settings=args)
     test_reader = imagenet_reader.val(settings=args)
 
-    train_data_loader.decorate_sample_list_generator(train_reader, place)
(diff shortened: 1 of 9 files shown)


PaddleCV/image_classification/train.py
```diff
@@ -171,20 +180,21 @@ def train(args):
     train_reader = imagenet_reader.train(settings=args)
     test_reader = imagenet_reader.val(settings=args)
 
-    train_py_reader.decorate_sample_list_generator(train_reader, place)
-    test_py_reader.decorate_sample_list_generator(test_reader, place)
+    train_data_loader.decorate_sample_list_generator(train_reader, place)
```
jerrywgz: set_sample_generator

shippingwang:
closed it cause same branch name conflict, already rewrite it in https://github.com/PaddlePaddle/models/pull/3660

PR #3660, opened 2019-10-18
Update api in image classification 

shippingwang:
- add PaddlePaddle version checking function
- update new data loader instead of py_reader
- use fluid.data interface instead of fluid.layers.data


jerrywgz (approved):
LGTM

PR #3668, opened 2019-10-18
fix spade infer for 1.6(#3575)

ceci3:
* fix spade

* update SPADE

* update spade

* update infer

* update infer


baiyfbupt (approved):
LGTM

PR #3711, opened 2019-10-22
simnet update to paddle v1.6

Dilyar93:
1. added download.py for windows users to download data set and pre-trained model.
2. update some interfaces in simnet for paddle v1.6.
3. fix some bug.


kuke (approved):
LGTM

PR #3719, opened 2019-10-22
[PaddleSlim] add skip_quant test in doc 

slf12:
add skip_quant test in doc

### DIFF
--- a/PaddleSlim/classification/quantization/compress.py
+++ b/PaddleSlim/classification/quantization/compress.py
@@ -100,14 +100,14 @@ def compress(args):
     com_pass.config(args.config_file)
     com_pass.run()
     conv_op_num = 0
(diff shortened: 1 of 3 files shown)


PaddleSlim/classification/quantization/compress.py
```diff
@@ -99,6 +99,16 @@ def if_exist(var):
         distiller_optimizer=None)
     com_pass.config(args.config_file)
     com_pass.run()
+    conv_op_num = 0
+    fake_quant_op_name = 0
```
wanghaoshuang: name?

slf12: done

wanghaoshuang (approved):

PR #3740, opened 2019-10-24
Fix bug of BMN/BSN, modify reader to match windows and refine the code for prediction/evaluation

huangjun12:
fix bug of bmn/bsn from QA;
modify reader to match windows;
refine code for prediction and evaluation in terms of outputs addition.


SunGaofeng (approved):
LGTM

PR #3791, opened 2019-10-28
fix bug of tall when using python3

huangjun12:
difference between python2 and python3 in the following items:
string and byte object;
map to list


SunGaofeng (approved):
approved

SunGaofeng (approved):
approved

PR #3802, opened 2019-10-29
fix raise statement in python3

huangjun12:
difference of python2 and python3 in raise statement;
modify README in dataset/bmn


SunGaofeng (approved):
approved

PR #3825, opened 2019-10-30
add log out for tall reader

huangjun12:
add log out for tall reader, as data processing takes too much time


SunGaofeng (approved):
approved

PR #4063, opened 2019-12-12
Dv3: use set_value to set weight, fix Embedding data layout

iclementine:
1. use set_value() to set the value of PositionEmbedding instead of using ivar;
2. fix input data layout for all inputs of Embedding layer, the tailing sized-1 dimension is removed in all cases.
3. updata README, use develop version of paddle.
4. update documentation for all the inputs of Embedding layers.
5. fix tensorboard tags
6. use weight instead of the protected `_w` in Embedding(a sublayer inside PositionEmbedding).


lfchener (approved):
LGTM

PR #4070, opened 2019-12-13
modify senta dygraph

JesseyXujin:
1. add infer benchmark
2.modify network gru/bow
3.add more information in README.md


xyzhou-puck (commented):
please double-check the implementation of CNN, I believe the shape of the CNN output may not be matched with the static graph version since there is a padding in the CNN.

xyzhou-puck:
I check the GRU net, it is consistent with the static graph version.

xyzhou-puck (approved):
LGTM.

PR #4077, opened 2019-12-16
refine infer in image classification

shippingwang:
- fix bug
use args: test_batch_size to make test data
- exclude FC layers params and bias when finetuning different classes(!=1000) model
**NOTE: should rename all the models' last FC layers as 'final_fc_offset' and 'final_fc_weights'(WIP)**
- save the output as JSON file when doing infer and eval
- print and save real image_id, like:

image


PaddleCV/image_classification/reader.py
```diff
@@ -304,12 +316,15 @@ def read_file_list():
                     full_lines = []
                     for i in range(settings.same_feed):
                         full_lines.append(temp_file)
-                for line in full_lines:
+
+                for line in full_lines[0:64]:
```
LielinJiang: why 0->64?

shippingwang: OK updated

heavengate (approved):

PR #4095, opened 2019-12-19
support out json in multi-card eval and add exclude fc params in finetune

shippingwang:
- support output the JSON file while doing multi-card eval in the image classification
**Note:**  To solve the problem about the framework cannot pass string variable into the network and there is a strong demand to get  it, we create a reader which always generate a batch of image_id(string) and image_data(data and label) and only feed the image_data when the executor runs, the python interface: **feeder.feed_parallel** is workable to implement it, note that **feeder.decorate_reader** ask the reader should be consistent with the feeder
After doing this, the exe.run will return the observation value in the fetch list as a shape of [the single card batch size, the size of fetch_list, image_shape]
The JSON file will be saved at the ./ in default
- add args:exclude to load pre-trained model correctly while doing finetune
- refine some warning

### DIFF
--- a/PaddleCV/image_classification/infer.py
+++ b/PaddleCV/image_classification/infer.py
@@ -101,8 +101,8 @@ def infer(args):
     place = fluid.CUDAPlace(0) if args.use_gpu else fluid.CPUPlace()
     exe = fluid.Executor(place)
     exe.run(fluid.default_startup_program())
-
-    places = fluid.framework.cuda_places()
+    if args.use_gpu:
+        places = fluid.framework.cuda_places()
     compiled_program = fluid.compiler.CompiledProgram(
         test_program).with_data_parallel(places=places)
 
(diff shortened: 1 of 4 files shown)


PaddleCV/image_classification/infer.py
```diff
@@ -101,8 +101,8 @@ def infer(args):
     place = fluid.CUDAPlace(0) if args.use_gpu else fluid.CPUPlace()
     exe = fluid.Executor(place)
     exe.run(fluid.default_startup_program())
-
-    places = fluid.framework.cuda_places()
+    if args.use_gpu:
+        places = fluid.framework.cuda_places()
     compiled_program = fluid.compiler.CompiledProgram(
         test_program).with_data_parallel(places=places)
```
LielinJiang, line 108: 如果use_gpu是false的话，这里是不是就报错了。

shippingwang: updated

heavengate (approved):

PR #4096, opened 2019-12-20
fix doc for post_training_quantization, test=develop

juncaipeng:
fix doc for post_training_quantization, test=develop


wzzju (approved):
LGTM.

PR #4118, opened 2019-12-24
update README and fix random nan in rcnn startup

heavengate:
**update README and fix random nan in rcnn startup**
- change conv2d weight initializer from xavier to kaiming for avoiding random nan on RCNN startup
- fix typo and update README


jerrywgz (approved):
LGTM

PR #4120, opened 2019-12-24
fix a bug in language model

lfchener:
fix a bug in language model that has no attribute 'get_cuda_device_count' when use CPUPlace.


iclementine (approved):
LGTM~

PR #4128, opened 2019-12-25
fix mnist multi gpu save error

phlrain:
fix mnist multi gpu save error


DDDivano (approved):
LGTM

DDDivano:
Fix mnist multi gpu save error

PR #4133, opened 2019-12-26
fix bugs in classification

shippingwang:
- reduce reader_buffer size of xmap_readers in image classification 2048->64
- pass args:warm_up_epochs in func: cosine_warmup_decay

### DIFF
--- a/PaddleCV/image_classification/utils/optimizer.py
+++ b/PaddleCV/image_classification/utils/optimizer.py
@@ -37,7 +37,10 @@ def cosine_decay(learning_rate, step_each_epoch, epochs=120):
     return decayed_lr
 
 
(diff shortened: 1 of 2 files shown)


PaddleCV/image_classification/utils/optimizer.py
```diff
@@ -37,7 +37,10 @@ def cosine_decay(learning_rate, step_each_epoch, epochs=120):
     return decayed_lr
 
 
-def cosine_decay_with_warmup(learning_rate, step_each_epoch, epochs=120):
+def cosine_decay_with_warmup(learning_rate,
+                             step_each_epoch,
+                             warm_up_epoch=5.0,
```
qingqing01, line 42: warm_up_epoch -> warmup_epochs

Issue #4110, opened 2019-12-23
BMN/BSN模型评估时报错


li-900406:
错误如下：
Process Process-13:
Traceback (most recent call last):
  File "D:\Anaconda\lib\multiprocessing\process.py", line 297, in _bootstrap
    self.run()
  File "D:\Anaconda\lib\multiprocessing\process.py", line 99, in run
    self._target(*self._args, **self._kwargs
  File "G:\models-develop\PaddleCV\PaddleVideo\models\bsn\bsn_utils.py", line 124, in video_process
    result_dict[video_name[2:]] = proposal_list
NameError: name 'result_dict' is not defined
[INFO: metrics_util.py:  420]: [EVAL] eval finished.    Loss = nan
W1223 18:39:51.984580 17320 device_context.cc:236] Please NOTE: device: 0, CUDA Capability: 61, Driver API Version: 9.2, Runtime API Version: 9.0
W1223 18:39:51.997812 17320 device_context.cc:244] device: 0, cuDNN Version: 7.6.

请问该怎么解决，谢谢！
![捕获](https://user-images.githubusercontent.com/59164417/71354279-665f0400-25b6-11ea-95b9-abab41c2abce.PNG)

wilhelmzh:
辛苦 @huangjun12 帮忙看下

huangjun12:
result_dict是在bsn_utils.py文件的bsn_post_processing函数中定义的global变量，报错信息提示未找到该变量。为方便定位问题，建议贴一下配置环境、执行的命令及所有的报错信息~  
另外BSN模型分TEM、PGM和PEM三个模块，各模块执行顺序有要求，详情参考 https://github.com/PaddlePaddle/models/blob/develop/PaddleCV/PaddleVideo/models/bsn/README.md

li-900406:
环境为 windows10 cuda 9.0 cudnn7.6 paddlepaddle是在官网下载的对应版本
执行命令为 bash run.sh predict BMN ./configs/bmn.yaml
结果如下
![捕获1](https://user-images.githubusercontent.com/59164417/71406882-6cb3b580-2674-11ea-932d-2dbb20c39e19.PNG)
![捕获2](https://user-images.githubusercontent.com/59164417/71406884-6de4e280-2674-11ea-9d4d-04557f2fadc5.PNG)

huangjun12:
[...]

PR #4134, opened 2019-12-27
fix bmn/bsn bug caused by the global variable in windows

huangjun12:
fix bmn/bsn bug caused by the global variable of multiprocessing  in windows. 
Please refer to https://github.com/PaddlePaddle/models/issues/4110


SunGaofeng (approved):
LGTM

PR #4142, opened 2019-12-30
fix PointNet++ hang on force exit.

heavengate:
**fix PointNet++ hang on force exit.**
PyReader multi-thread not exit properly on CTRL+C force exit, kill process group on force exit.


jerrywgz (approved):
LGTM

PR #4154, opened 2020-01-04
Optimizer init with parameters, test=develop

JepsonWong:
There is an incompatible API change for Optimizer, \_\_init\_\_ in dygraph must have a parameter_list arg. See https://github.com/PaddlePaddle/Paddle/pull/20815 .

This PR upgrades the ptb_lm model.


zhiqiu (approved):
LGTM

PR #4191, opened 2020-01-13
dygraph_seq2seq_fix_infer

1024er:
fix infer process for v1.7


guoshengCS (approved):

peterzsj6:
Hi.1024er. I am trying to implement the seq2seq dygraph demo on colab right now. But I still can't infer using the new code(getting single output instead of a sequence).

peterzsj6:
(About the reader module in the dygraph seq2seq model）
Sorry to trouble you this way, but could you explain a little bit about why are we using ids = [1] + ids + [2] in the tar data language? Since [1][2] has already been assigned in vocab, I am a little confused as what [1],[2] means in this case.


Also, I posted an issue under the paddle model repository, the reader module  seems to have a typo. (def read_all_line(filenam):)


Much appreciated!

PR #4206, opened 2020-01-16
Update dygraph Transformer to be consistent with its graph counterpart.

guoshengCS:
Update dygraph Transformer to be consistent with its graph counterpart.


xyzhou-puck (approved):
LGTM.

PR #4255, opened 2020-02-10
Add Dygraph BERT model

xyzhou-puck:
Add the dygraph BERT model (version 1).

I test it on glue (mainly in MNLI and CoLA) and  I got similar results to its static-graph version.

I will add source code of SQuAD and pre-training later.


kuke (approved):
LGTM

PR #4297, opened 2020-02-17
Update resnet_acnet.py

parap1uie-s:
Bias attr of square conv should be "False" rather than None during training mode.


littletomatodonkey (approved):
LGTM

shippingwang (approved):

PR #4308, opened 2020-02-18
Refine api of emotion detection and dialogue domain classification

xyzhou-puck:
Refactor emotion detection and dialogue domain classification to make them fit paddle 1.7.


kuke (approved):
LGTM

PR #4319, opened 2020-02-20
upgrade save and load interface (#4311)

xjqbest:
* upgrade dcn and xdeepfm, change from all old save/load api to fluid.save and fluid.load
* test=develop


123malin (approved):
LGTM

PR #4329, opened 2020-02-21
Make dygraph Transformer support to load py2 saved models.

guoshengCS:
Make dygraph Transformer support to load py2 saved models.


xyzhou-puck (approved):
LGTM.

PR #4337, opened 2020-02-25
fix fluid.io.save/load to fluid.save/load

SunGaofeng:
1. change fluid.io.save/load to fluid.save/load
2. modify related codes that load params specifically, such as Non-Local and StNet models.
3. change the parameters' format in the download links and related documents.

### DIFF
--- a/PaddleCV/PaddleVideo/utils/train_utils.py
+++ b/PaddleCV/PaddleVideo/utils/train_utils.py
@@ -123,7 +123,7 @@ def train_with_dataloader(exe, train_prog, compiled_train_prog, train_dataloader
                                  test_fetch_list, test_metrics, log_interval,
                                  save_model_name)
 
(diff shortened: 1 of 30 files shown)


PaddleCV/PaddleVideo/utils/train_utils.py
```diff
@@ -115,40 +115,15 @@ def train_with_dataloader(exe, train_prog, compiled_train_prog, train_dataloader
...
-        '_final',
-        save_type='.pdckpt')
-    save_model(
-        exe,
-        train_prog,
-        save_dir,
-        save_model_name,
-        '_final',
-        save_type='.pdparams')
+    save_model(exe, train_prog, save_dir, save_model_name, '_final')
```
shippingwang: 所以现在不加_final了么？ 那这里还是保存着_final ?

qingqing01 (approved):

PR #4394, opened 2020-03-10
Fix Transformer path in README

guoshengCS:
1. Fix Transformer path in README
2. Fix the leaving gen_data.sh in dygraph Transformer.


kuke (approved):
LGTM

PR #4164, opened 2020-01-06
remove init_on_cpu from models

zhangting2020:
`init_on_cpu` is used to force the variable to be inited on CPU or force the CPUkernel is executed. Variables related to the learning rate do not participate calculations in model, so there is no need to calculate on the GPU. However, the usage of `init_on_cpu` in the models does not take effect because it only works for some Ops with force_cpu Attr. 

test case 1:
```python
import paddle.fluid as fluid
from paddle.fluid.initializer import init_on_cpu
from paddle.fluid.layers.learning_rate_scheduler import _decay_step_counter

LEARNING_RATE = 0.01
TOTAL_STEP = 30000
POWER = 0.9

def poly_decay():
    global_step = _decay_step_counter()
    with init_on_cpu():
        decayed_lr = LEARNING_RATE * (fluid.layers.pow(
            (1 - global_step / TOTAL_STEP), POWER))
    return decayed_lr

lr = poly_decay()
place = fluid.CUDAPlace(0)
exe = fluid.Executor(place)
exe.run(fluid.default_startup_program())
result = exe.run(fetch_list=[lr])
```
log1
In runtime, CUDAKernel for pow Op is executed.
```
I0106 08:42:36.204762  7559 operator.cc:975] CUDAPlace(0) Op(scale), inputs:{ScaleTensor[], X[tmp_0:float[1]({})]}, outputs:{Out[tmp_1:float[1]({})]}.
I0106 08:42:36.204845  7559 executor_gc_helper.cc:166] Erase variable tmp_0
I0106 08:42:36.204968  7559 operator.cc:1060] expected_kernel_key:data_type[float]:data_layout[ANY_LAYOUT]:place[CUDAPlace(0)]:library_type[PLAIN]
I0106 08:42:36.205127  7559 operator.cc:975] CUDAPlace(0) Op(pow), inputs:{FactorTensor[], X[tmp_1:float[1]({})]}, outputs:{Out[pow_0.tmp_0:float[1]({})]}.
I0106 08:42:36.205235  7559 executor_gc_helper.cc:166] Erase variable tmp_1
```

test case 2:
```python
import math
import paddle.fluid as fluid
import paddle.fluid.layers.ops as ops
from paddle.fluid.initializer import init_on_cpu
from paddle.fluid.layers.learning_rate_scheduler import _decay_step_counter

def cosine_decay_with_warmup(learning_rate, step_each_epoch, epochs=120):
    """Applies cosine decay to the learning rate.
[...]

PR #4444, opened 2020-03-18
remove init_on_cpu

zhangting2020:
The usage of `init_on_cpu` does not take effect because it only works for some Ops with `force_cpu` Attr. For more information, please refer to this [PR](https://github.com/PaddlePaddle/models/pull/4164) in models repo.

And this API will be deprecated in the future.


chajchaj (approved):
rm init_on_cpu

huangjun12 (approved):
LGTM

PR #4592, opened 2020-05-07
Fix dygraph cyclegan

ceci3:
Fix dygraph cyclegan
add img_with_label data reader

polish data_reader according to https://github.com/PaddlePaddle/models/issues/4587


tink2123 (approved):
LGTM

PR #4620, opened 2020-05-14
update .batch and .shuffle apis

1024er:
update .batch and .shuffle apis for PaddleNLP/lexical_analysis

paddle.batch -> fluid.io.batch
paddle.reader.shuffle -> fluid.io.shuffle


kuke (approved):
LGTM

PR #4628, opened 2020-05-14
Add greedy search for dygraph Transformer.

guoshengCS:
Add validation for dygraph Transformer.
Add cross-attention cache for dygraph Transformer.
Add greedy search for dygraph Transformer.

Re: #4233


kuke (approved):
LGTM

PR #4652, opened 2020-05-19
Added an option to choose to run sentiment classification for CPU and selected model

jczaja:
This PR introduce capability to choose to run  sentiment classification training/inference on CPU and also with selected model without modifying the script. It will be useful for us to run automatically sentiment classification training for CPU.

By default run.sh works as before , but it is possible to instruct from commandline to run for CPU and use bow_net :
`USE_CUDA=false MODEL_TYPE=bow_net run.sh train`


lidanqing-vv (approved):
LGTM.

kuke (approved):
LGTM

PR #4953, opened 2020-11-15
fix tsn benchmark

shippingwang:
Changes:
In norm function: Attribute: loc and sth. change to std and mean, they are back now.
paddle.load only deal with one kind of inputs, either opt parms or weight params, BIG improvement!
almost forget this: PiecewiseLR is named to PiecewiseDecay, again, so philosophy.
It is a sort of meaningful PR, really, our semi-auto benchmark platform will smoothy run now except for another API changes in the future :)


ceci3 (approved):
LGTM

PR #4974, opened 2020-11-26
Enabled CPU,  time logging and args for se_resnext Dygraph

arlesniak:
Enabled CPU and added more robust logging and args for se_resnext Dygraph, print args, max_iter, time added


jczaja (approved):
LGTM

PR #5011, opened 2020-12-10
Unified the task name of DGU with paddle1.8

xiemoyuan:
Unified the task name of DGU with paddle1.8, and update the doc.


ZeyuChen (approved):
LGTM

PR #5045, opened 2020-12-14
clean hapi notes, clean eager_run and modify num_labels to num_classes

jeff41404:
1. clean hapi notes
2. clean eager_run
3. modify eletra modeling num_labels to num_classes, same with bert


ZeyuChen (approved):
LGTM

PR #5060, opened 2020-12-15
Add 2 datasets and update squad metric and run_du

smallv0221:
Add 2 datasets and update squad metric and run_du


ZeyuChen (approved):
LGTM

PR #5113, opened 2020-12-18
improve_example_readme

jeff41404:
1. imporve_example_readme, modify full name and description of electra
2. delete electra/run_glue.py，unified use glue/run_glue.py


ZeyuChen (approved):
LGTM

PR #5117, opened 2020-12-18
fix bert benchmark fleet compiled

wangxicoding:
Fleet minimize will compile the program, if called `build_compiled_program`,  the program will be recompile, and distributed information will loss.


ZeyuChen (approved):
LGTM

PR #5173, opened 2021-01-04
Support xpu for ernie and bert in pre-train/glue task

wawltor:
Support xpu for ernie and bert in pre-train/glue task


ZeyuChen (approved):
LGTM

PR #5175, opened 2021-01-05
Support pad sequence and batch size can be multiplied by a factor

FrostML:
Transformer support pad sequence and batch size can be multiplied by a factor


guoshengCS (approved):

FrostML:
【待数据压缩包上传后合入】

PR #5211, opened 2021-01-18
Update vocab.py

fiyen:
fix some errors in to_json and from_json.


ZeyuChen:
Since we have move PaddleNLP to https://github.com/PaddlePaddle/PaddleNLP . You can swift your PR to that repo if possible. Thanks for your contribution!

PR #5215, opened 2021-01-20
update codes for static and dynamic graph unification

Steffy-zxf:
update codes for static and dynamic graph unification


wawltor (approved):
LGTM

PR #5263, opened 2021-02-02
support padding direction selection

yingyibiao:
add an optional boolean argument pad_rght to Pad class to support padding direction selection


ZeyuChen (approved):
LGTM

PR #5278, opened 2021-02-10
fix a bug in fp16 training of BERT model that will cause non-converge…

huangxu96:
fix a bug in fp16 training of BERT model that will cause non-converge.


ZeyuChen:
This PR should move to https://github.com/PaddlePaddle/PaddleNLP this repo.

PR #5311, opened 2021-05-18
PaddleAudio project refactoring

ranchlai:
Project refactoring, major changes include:  
1.  Feature/backend api changed to align with librosa
2. Unit test added for feature and backends 
3. Audioset training example added with mAP0.401


ZeyuChen (approved):
LGTM

PR #5326, opened 2021-07-19
 Export parameters for setting power_to_db in LogMel transform

ranchlai:
1. export parameters for power_to_db in LogMel transform 
2. default to_db to False to be consistent with functional
3. fixed typo in docstring


ZeyuChen (approved):
LGTM

PR #5330, opened 2021-07-21
update audioset example to use new functionals

ranchlai:
update audioset example to use new functionals
all test passed.


ZeyuChen (approved):
LGTM

PR #5334, opened 2021-08-05
Transforms and functionals update

ranchlai:
Updated stft、melspectrogram in functionals and transforms. Now features are compatible with librosa up to 1e-7 mean error. 
Major changes are: 
a) dtype in window/functional/transforms are controllable 
b) Export htk and norm parameters to be compatible with librosa and torchaudio.
c) Remove NoiceSource  and RIRSource and leave them to the future speaker example, since they require reading data from disk and thus not suitable as transforms.
d) Update docstring accordingly.
e) Test updated.


ZeyuChen (approved):
LGTM

PR #5641, opened 2022-11-22
update Ernie-Layout

doubleguy:
The app folder has been deleted because the key used in it is not public


D-DanielYang:
invalid pr, closed

PR #5728, opened 2023-02-27
Fix url to download raw code

Yam0214:
Replacing the error URL, and using `wget` to download raw code correctly.


chenxiaozeng (approved):
LGTM

